PR #1, opened 2016-08-30
the quests outside the moradan can be called from quest helper

onurcanbektas:
-  2. part of the lover letter quest in the 1.lua but there is a problem with checking the state of the love letter (56) quest, Royal Guard sees the status of the quest  0 all the time, I don't know why.


srmeier:
Quest state requires that rows be added to the QUEST_HELPER table for the state of a quest to switch from 0 to 1 or 2. I'm planning to post a video either tomorrow or the day after detailing this. But that's why the Royal Guard sees the status of the quest as 0.

srmeier:
I wish there was a way to automatically populate all the rows in the QUEST_HELPER table. The issue with writing a program (similar to the EvtConverter) to do that is that you require the NPC SID which isn't available in the EVTs. There could be some other way of automating this though, I'm not sure. Currently you have to add a row to the quest helper table for each quest state possible for a particular NPC.

onurcanbektas:
no, I added those to the QUEST_HELPER table but neither in love letter quest nor in instrument quest in Royal guard checking the state of a quest correctly is unsuccessful.

srmeier:
Humm, interesting. I'll take a look and see if I can figure out what's going on.

twostars:
This issue was moved to ApexKO/OpenKO#1

PR #2, opened 2016-09-04
Clan processes and adding friend is handled.

onurcanbektas:
But I've made some chances in the database too,  so without them it
probably won't work.Plus , I added a features.txt so that we can easily
see what is working and what is not working, and the tests we've made,
which allows other people to contribute to the project even by just
testing the system.


srmeier:
Everything looks good here. In User.cpp you add a WIZ_NPC_SAY flag, I don't believe the client ever sends that packet but I could be wrong about this. What made you add the WIZ_NPC_SAY flag to the switch-statement?

srmeier:
Also, when you say that you made changes to the DB, what sort of changes are you talking about?

onurcanbektas:
when you reach the end of the conversation with the npc, after pressing the OK button, source client sends the the packet with the number of 56 and it shows in the output tab of the VS.I just added there, so that it doesn't cause any confusion.

onurcanbektas:
In database, I give knight user to the permissions like execute, insert, update etc. on KN_online.Moreover, I chanced 2 stored procedure. I will send images of them on skype.

twostars:
This issue was moved to ApexKO/OpenKO#2

PR #3, opened 2016-09-09
There is some chances in Upgrade system, CURRENTUSER update and SearchQuest lua script.

onurcanbektas:
Note that, I editted some of the parts in 2.lua and 1.lua because there was some problem in the conversion of the code.

By the way, if it will be more comfortable, I can create these pull requests to the develop branch.


twostars:
This issue was moved to ApexKO/OpenKO#3

PR #9, opened 2016-10-27
Added more IDS_TAGS

ForcePower:


GiladRe:
Hey there :) i just realized what you wrote here: 

> Our mission now is fixing nd adding the right IDS Tags.
> We are doing it because we removed the windows app which were used this
> strcture IDS_XX_XX string and pull it from the client and this because
> we want the client to work as well on Linux & Mac , and while in your
> code your dependce on Windows app you can't achieve this goal.

Sorry i haven't seen that. in my opinion we can stick to the string constants, because all the translation would be in one place and in different languages. So it would be much easier for us to translate instead of traveling each file and changing those constants to strings. 
About making it work on cross platforms, don't worry about it, it shouldn't be a big trouble getting that table work on cross platforms, because we could take that table and turn it into an XML file that would be loaded and stored into cache once the client started. So we don't get a performance hit that way.
This is how string tables in VS works, it's allocating a place in memory and stores in cache and as soon as you calling a constant string, it loads from the binary with his id and looking for the text in the resource script. It's not like it needs to load the whole table for each string constants. The system is intelligent enough to do those calculation. 

Alternative: We could also create a enum List object and simply put all the strings direct in code, but that could be a performance hit, so the best option would be XML or a Table file inside Data and load it once client started.

Btw, i created a localization-translate branch for this purpose in case we going to stick the string table constants so it doesn't affect the master branch code until it's done.

It's good to see people getting interested and willing to contribute! kind of motivates me to dig into KO again even though i barely have time due to work and studies. But i keep reading the commits in this project and trying to follow up you guys :)

srmeier:
Thanks Gilad, you bring up some good points. I honestly don't see up using a different language besides english in the future. One benefit of working with KO so late in its lifetime is that you get to see what people are most interested in. I'd have to day that the english language has been most popular by far. In addition, if we decide to support new languages in the future the real challenge won't be with strings as so much updating GUI textures for all the UI in the game, which I don't see happening. Support for the Turkish language is perhaps the most likely but even then most of the Turkish players in KO are use to the english language, so I'm not sure.

I think for now we'll drop the multi-language support to ease our ability to port between different operating systems. As long as we have the IDs and their corresponding strings (both we can search for via the source code assuming the IDs remain in the source but simply commented out) I think we'll be able to easily revisit this issue in the future. For now I'll revert the commits and continue in this direction but I agree that this is a potential issue. If we had a large community of people on the forums I'd say let's pose the question for multi-language support and let them decide (listing pros and cons), but I'm not sure enough people visit codingbackwards.in at the moment.

PS - You should jump on Skype more! ;P

srmeier:
Actually before I revert this. Could you jump on Skype at some point and let's discuss it. I see you have done a lot here and I want to understand what you had in mind. You, me, and FowerPower should all jump in a Skype call sometime in the next few days.

srmeier:
I should also mention that while creating new branches for this kinda stuff is ideally what you'd want to do, not enough people are using this code live so the master branch is basically our be-all end-all development branch haha

GiladRe:
Thanks for the input Stephen :) 
You are totally right about that English is the most used language for our case, but having strings directly in code that displays to the client is a bad practice to be honest. It's not a big thing though, but having everything in one place, is always nice to have as long as it doesn't cause any issues to the project of course and makes the project more flexible to modify or extending. 
It's even faster to translate that way, imagine in the future you'll want to extend the user interface to have another language, you simply create another string table since we already got a resource file and translate it directly into that table and let the code structure do his job. 

I created that  branch in case you would want to revert this so it won't affect the main project. I find branches much easier to work, but of course we can stick to the pull requests GitHub interface. 
>  so the master branch is basically our be-all end-all development branch haha
Haha... sounds like a deal  :P 

Let's discuss it in Skype and catch up a bit ;) Probably on the weekend or so if not today.

onurcanbektas:
As a Turkish person, I agree about what Stephen said, most Turkish players prefer to play it in English.If it what they are used to.

GiladRe:
> As a Turkish person, I agree about what Stephen said, most Turkish players prefer to play it in English.If it what they are used to.

Hey there :) Yea, i totally agree with him too. Though the client have many messages strings that would be good translated to Turkish. When you have tables, you get the flexibility modifying it the way you want and let the user decide in options what language he wants to pick.

onurcanbektas:
> Hey there :) Yea, i totally agree with him too. Though the client have many messages strings that would be good translated to Turkish. When you have tables, you get the flexibility modifying it the way you want and let the user decide in options what language he wants to pick.

Don't get wrong, I totally agree with you in that idea, particularly 

> the best option would be XML or a Table file inside Data and load it once client started.

srmeier:
For the moment let's keep the new branch and revert the master so I can merge Force's commits. Sometime today or tomorrow let's have a Skype call (you're invite too Onurcanbektas if you'd like).

Gilad I sent you a message on Skype, get back to me when you can.

onurcanbektas:
Due to time difference, I don't think I will make it, but when I get home, I will be online in Skype.

twostars:
This issue was moved to ApexKO/OpenKO#9

PR #22, opened 2017-01-20
EvtConverter is updated

onurcanbektas:


srmeier:
This is a very solid update! Very nice work!

twostars:
This issue was moved to ApexKO/OpenKO#22

PR #23, opened 2017-01-20
Evtconverter is updated and some lua functions are added to the server side

onurcanbektas:


twostars:
This issue was moved to ApexKO/OpenKO#23

PR #27, opened 2017-02-22
Onurcan has started adding support for Unix threads

srmeier:


twostars:
This issue was moved to ApexKO/OpenKO#27

PR #28, opened 2017-02-22
Mutexes are done! but, the type of thread functions haven't been handled completely.

onurcanbektas:


srmeier:
Can you move the mutex definitions (i.e. calls to "malloc()") to the constructors (and "free()" to destructor) and leave the mutex declarations in the class declarations? If this is confusing let me know and I'll point out specifically what I mean.

srmeier:
Also some of the header files were commented out from stdafx and the #define IsWinDef you had there wasn't working for some reason. May need to find another way to do that.

onurcanbektas:
Sure, but free() is already in the destructor of the class.By the way, what do you mean by "leave mutex declarations in the class declarations" ? 

About the commented header files, since I am not able to fully compile the code, I need need to use the some part of the code in order to test, and I forgot to undo that before pushing the commit, but in this commits, I corrected that.

onurcanbektas:
By the way, was #define IsWinDef working in the first pull request ?

srmeier:
basically I mean leave `Mutex* m_pMutex;` in the class declaration and put the malloc()s in the constructor.

`#define IsWinDef` wasn't working in the first pull request. I had to add a commit which changed it around for the time being. I've never see it done that way either, we may have to think of another way to do it. I also think you'll need to pull that most recent commit too. I don't think that'll be a problem though because I didn't change much of anything.

onurcanbektas:
I had considered to allocate the memory for m_target member, but in that way, it was giving segmentation fault, so that was the reason why I used the malloc outside of the class, but I'm going to check again whether there is some way to do that.

P.S, The conflict has resolved.

srmeier:
Ok sounds good, I'll take a look at this in the morning and merge/check for bugs.

onurcanbektas:
By the way, I pushed a new commit, which needs testing on windows about whether return value by the thread function is returned properly.I only implemented it to TimeThread() function in TimeThread.cpp, so you can use that.

srmeier:
The return value seemed to work fine although, no idea why, but the other server projects weren't able to local the RETURN_THREAD() at link time. I moved this function from stdafx.cpp to TimeThread.cpp and it seemed to resolve the issue.

When do you plan to get a windows machine?

onurcanbektas:
As far as I know, never :). I mean in campus, for linux, I can even use the computers in labs, but there is no way that I will be able to have a windows machine which I can use for the project.

onurcanbektas:
I think someone except than you should also test these last commits of mine because you are getting weird errors lately :)

srmeier:
Haha yeah, and it would take more work off me - which is nice.

twostars:
This issue was moved to ApexKO/OpenKO#28

PR #31, opened 2017-04-01
WIZ_TIMENOTIFY disconnect in Release mode.

Demircivi:
Server doesn't allow user to send unhandled packets in Release mode,
since the Client sends WIZ_TIMENOTIFY packet periodically, we have to
allow it in Server side.


twostars:
This issue was moved to ApexKO/OpenKO#31

PR #32, opened 2017-04-02
Fix for Clan page.

srmeier:


twostars:
This issue was moved to ApexKO/OpenKO#32

PR #49, opened 2017-04-05
Added unhandled packet notifier

Demircivi:
When an unhandled packet received, player will see a message in
ChatWindow. This can be shown in Debug builds only.
Also fixed some tabbings. Nothing special.


srmeier:
Unhandled packets in the ChatWindow for debug builds is smart

twostars:
This issue was moved to ApexKO/OpenKO#49

PR #53, opened 2017-05-18
Misc changes

twostars:
Not the most ideal pull request, but since this is temp I figured doing it this way would save time / keep things synced neater.


twostars:
This issue was moved to ApexKO/OpenKO#53

PR #61, opened 2017-05-27
Floating numbers added to player experience bar

cbagriacik:


twostars:
Try to avoid reformatting the entire file. It makes it hard to see the actual changes, since the diff shows the entire file was changed.

cbagriacik:
Sorry about that, i noticed after commit. But Visual Studio showed me only
changed lines before commit not whole file.

2017-05-29 11:32 GMT+03:00 twostars <notifications@github.com>:

PR #62, opened 2017-06-19
Makefiles are added, and a couple of minor problems are handled.

onurcanbektas:
Figuring out how to link the files correctly and that inline problem took my 2 days, so please be kind to those files :)


srmeier:
Is "Server/AIServer/main" the binary file you are producing? We should probably try and remove that before I merge these files. Other than that everything looks good.

onurcanbektas:
Yes, that is the executable, but it is deleted already. I was probably in the repo because I forgot to delete before, but right now it is deleted.

srmeier:
Gotca!

PR #64, opened 2017-07-30
There was a ; missing that stopped building shared.lib

DoogeJ:


srmeier:
Good catch! Thanks for submitting these @DoogeJ

DoogeJ:
Np! Trying to figure out how to use the client that I finally managed to build ;) will watch your video's on it soon!

Thanks for all your work! <3

PR #67, opened 2017-08-23
Notification is added when the player receives an item 

onurcanbektas:


onurcanbektas:
ekran goruntusu 4

onurcanbektas:
@twostar I tested it and yes, if, for example, after completion of a quest, npc take item from you, client again says "You've received blah item.". I will correct it, hopefully today.

PR #69, opened 2017-08-24
My works on this project. I think we should add data files here to track changes on them. Data files may need to change in time and client depends data files.

d5dl2:


srmeier:
The data files where originally in the repo but I decided to remove them because they were taking up way too much space. What sort of modifications to the data files are you suggesting? The idea I had in mind was that people should be able to just use the stock 1298 data files.

onurcanbektas:
Maybe we can create another repo for the data files ?

onurcanbektas:
By the way, @d5dl2 you don't need to leave your signature to every code that you have written, github already does that job.

d5dl2:
Ok, thanks. In character create screen, informations of classes and races was doesnt work. I fix it on 
59da83c commit. In data files, Karus Puri Tuarek information was in Korean Lang. I translated it to English on Text_us.tbl. Also i changed nation select screen data files (Co_nationselect_us.uif and some dxt's..) from another data files. Because nation select screen was not worked and i was cant find the way to fix it. With these commits, nation select and character create screens works pretty well. I can add images if you need it.

srmeier:
Ok sounds good. And we can add another repo for data files but I think sticking with the default 1298 files works just as well.

Here are some links to data files:
http://stephenmeier.net/files/data.zip
http://stephenmeier.net/files/data_1298.zip

In case you don't have them already. I think the "data.zip" might actually be 1264. Those might be the data files I was originally using btw.

d5dl2:
Ok, ill check if data_1298.zip may work with orginal code or with changes. Ill try to not update data files unless it is really important.

________________________________
Gönderen: Stephen Meier <notifications@github.com>
Gönderildi: 25 Ağustos 2017 Cuma 16:44
Kime: srmeier/KnightOnline
Bilgi: d5dl2; Mention
Konu: Re: [srmeier/KnightOnline] My works on this project. I think we should add data files here to track changes on them. Data files may need to change in time and client depends data files. (#69)


Ok sounds good. And we can add another repo for data files but I think sticking with the default 1298 files works just as well.

Here are some links to data files:
http://stephenmeier.net/files/data.zip
http://stephenmeier.net/files/data_1298.zip

In case you don't have them already. I think the "data.zip" might actually be 1264. Those might be the data files I was originally using btw.

twostars:
Assets introduce repository bloat. Git was not designed for dealing with binaries.

There are ways of getting around this, but ultimately I agree with @srmeier; we should just be using the official 1.298 data files. Having to modify these data files remains outside of the scope of the project; the project is designed to support them as they are.

srmeier:
Just tested and this data.zip that I just uploaded works
http://stephenmeier.net/files/data.zip

PR #72, opened 2017-08-28
Added back in the ability to move camera when mouse goes to the edges…

SavvyNik:
Added back in some code that was commented out for allowing the camera to move when the mouse goes to the edges of the screen. I made a slight change to the code. I think this is a great feature to have. Was there a reason this was commented out? I tested it and it works well.

### DIFF
--- a/Client/WarFare/GameProcMain.cpp
+++ b/Client/WarFare/GameProcMain.cpp
@@ -1117,13 +1117,14 @@ void CGameProcMain::ProcessLocalInput(uint32_t dwMouseFlags)
 	//	s_pLocalInput->MouseSetPos(ptPrev_RB.x+x, ptPrev_RB.y+y);
 	//}
 
-	/*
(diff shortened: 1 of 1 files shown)


srmeier:
I believe I originally removed this because it doesn't work very well when in windowed mode (since the total screen coordinates change a little). Have you tested this in window mode? I don't mind adding this back and maybe even wrapping it in a if-statement that checks for whether GameBase::s_Windowed is set.

SavvyNik:
Yes, I've tested it in the windowed mode. It seemed to work pretty well for me and it works the way I remember it in the old KO. If you leave the window with the cursor you'll continue to move the camera until you get the cursor back inside the window again, but that's how it has always been. Maybe it'd be best to add it back in and see what other people think and then add the conditional for it's in windowed mode vs full screen.

srmeier:
Ok great! :+1:  Sorry I must be remembering wrong. Merged :)

PR #76, opened 2017-08-28
An issue has been solved, and a minor update

onurcanbektas:


twostars:
Yeah, no. There's no need to be modifying SDL.

The portability issue lies in using this structure directly in the first place, which we use because we're using DirectX directly.

Modifying SDL here like this doesn't solve anything; it merely causes more issues with compatibility (updating SDL becomes out of the question, when its headers are customised in this manner) and problems with defines overriding Windows' typedefs... more headaches down the track.

If you want to address this issue, it should be with moving away from needing to use DX, not hacking away at 3rd party libraries.

srmeier:
One minor point with this is that I believe you might have forgot to pull before you submitted this pull request because the "\" I added to some of the include files (had to help someone fix this earlier because it wouldn't let him compile) are missing.

twostars:
Closing this because it doesn't really address anything in particular. Also, the latter issue with HP showing in the target bar is resolved in the current branch.

onurcanbektas:
@srmeier Actually, I first pull from the repo, but there was some problems with VS2017, so I reverted it, but then I forgot to pull it again, lol.

onurcanbektas:
@twostars I'm new to SDL, so I will try to do it when I understand the system completely.

onurcanbektas:
@twostars  About the first option, I have search the documentations on SDL, but couldn't find anything about pure SDL implementation of windows messages, so could you give some direction or explain what you mean by "complete SDL implementation" ?

srmeier:
@onurcanbektas If you take a look here (https://github.com/srmeier/KnightOnline/blob/master/Client/WarFare/LocalInput.cpp) and search for "WndProcMain" - we need to eventually get rid of that, which means writing SDL implementations that achieve the same result as the WinAPI.

Also search for "SDL_SYSWMEVENT" in that cpp file. This is where SDL is passing the messages received from the WinAPI to the "WndProcMain" function.

PR #77, opened 2017-08-29
Camera now only turns if window has mouse focus.

SavvyNik:
Added a bool to check if the game window has mouse focus from an SDL event.  I also put conditional around the rotation of the camera in the case that a user moves to the edge of the screen with their mouse. Now the character will no longer rotate camera if they leave the game window with their mouse.


twostars:
Realistically this should extend to all mouse input, not just this specific behaviour but I'll merge this in. Thanks.

PR #82, opened 2017-08-30
The problem in the pull request $67 has been solved, hopefully.

onurcanbektas:
To make sure, I have tryied removing various items from character, and didn't saw any problem.

PR #85, opened 2017-08-31
Reverted the commits related with the pull request #67 and the code is edited according to the pull request #82

onurcanbektas:


twostars:
One slight thing before I merge it: just make sure that the item exists in the TBL and it's good.
We are still missing the durability/duration from the packet, which probably should be at the very least referenced so it can be added in properly but otherwise it's fine.

Client/WarFare/GameProcMain.cpp
```diff
@@ -3408,6 +3408,7 @@ void CGameProcMain::MsgRecv_ItemCountChange(Packet& pkt)		// Item Count Change..
 		{
 			char szBuf[256] = "";
 			__TABLE_ITEM_BASIC* pItem = CGameProcedure::s_pTbl_Items_Basic.Find(iID / 1000 * 1000);
+			if (pitem == nullptr) { return; }
 			std::string szFmt;  ::_LoadStringFromResource(IDS_ITEM_RECIVED, szFmt);
 			snprintf(szBuf, sizeof(szBuf), szFmt.c_str(), pItem->szName.c_str());
 			CGameProcedure::s_pProcMain->MsgOutput(szBuf, 0xff6565ff);
```
twostars, line 3414: Also something I meant to mention earlier regarding the previous changes:
This colour code is actually incorrect. It should be 0xFFFFFF00.

Furthermore, "CGameProcedure::s_pProcMain->" is redundant when we're already in CGameProcMain. It can just call MsgOutput() directly.

PR #87, opened 2017-08-31
GameDef.h ReStructure for 1298 & Stub out for NoviceClassChange

SavvyNik:
Updated the fx, ui_us, and zone structure to match the new tbl files from v1.298. I've also added all definitions from the ui_us.tbl into GameDef.h. 

I've also made a small stub out of implementing subopcode 6 for NoviceClassChange. I will implement a small function in the GameProcMain client side as described in the ase N3_SP_NOVICE_CLASS_CHANGE_REQ:

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -51,3 +51,4 @@ ItemEditor/Release/ItemEditor.lib
 Client/WarFare/Debug/gameprocmain.new.obj.enc
 Client/WarFare/Debug/gameprocmain.obj.enc
 Client/WarFare/Debug/Log.txt
+Client/dataxx/Chr/canon.n3anim
--- a/Client/WarFare/GameDef.h
+++ b/Client/WarFare/GameDef.h
@@ -566,143 +566,182 @@ typedef struct __TABLE_TEXTS
 // NOTE: temp 1298 zone structure
 typedef struct __TABLE_ZONE
 {
-	uint32_t       dwID;                 // zone ID
-	std::string	szTerrainFN;          // GTD
+	//24 Fields in the 1298 Zone.TBL structure
+	uint32_t		dwID;                 // zone ID
+	std::string		szTerrainFN;          // GTD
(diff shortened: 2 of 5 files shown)


twostars:
"I've also made a small stub out of implementing subopcode 6 for NoviceClassChange"
Note that it's used for all promotions to indicate a job/class change, not just this one case

twostars:
Blah, was lazy and used Github to merge those which screwed with line endings. Whatever.

SavvyNik:
@twostars "all promotions to indicate a job/class change" are you saying that this function should handle reskill & getting master skill, as well as, beginner skill quest? I also see a UIClassChange class that's already implemented in the client. Is this what was originally created to handle the class changes.

twostars:
@SavvyNik Hm. I forgot it had that odd half-assed UI. Yeah, the handler probably should go in there then even though it really, really doesn't need to, considering you're just forwarding stuff back anyway.
Like, it needs to access s_pPlayer (directly accessible from CGameProcMain), s_pFXMgr (again, directly accessible from CGameProcMain), CGameProcMain::m_pUIHotKeyDlg and CGameProcMain::m_pUISkillTreeDlg.

So to me, having that method in the other class seems like a waste of time (but most likely where mgame put it). It merely encapsulates it in the sole sense that it belongs to the class change related set of things (but it also has nothing to do with the UI, and accesses stuff outside of its purview).

With that reasoning I'd just keep the handler in CGameProcMain.

Also, I don't think reskills are handled by this specific UI (if that's what you mean), but both promotions are (101->105, 105->106 ). I've never really known what to call them, so in anything I've written I've referred to them as follows novice/beginner (101), skilled (105) and master (106), if that helps clarify things.

SavvyNik:
@twostars Yes, that clarifies things. I'll work on this tomorrow. Thanks for all the info.

PR #89, opened 2017-09-09
Few fixes and improvements.

### LINKED ISSUE #59: Issue with N3Sky not being large enough
So I believe a lot of people are familiar with this issue

pic00

Turns out if you just make the render billboard larger it fixes this problem.

pic01

Just opening and closing this issue so the solution gets documented.

GiladRe:
* Fixed broken sky issue that resolves #59 
* Fixed GM collision detection.
* Fixed /time command for gms.
* Implemented new /fog command for gms(useful for debugging things related to rendering and can be good for gms sometimes to play around with it). Note regarding this, that i added a flag that set to false since we don't want to affect developers who haven't update their Text_us.tbl and have some weird behavior.


janekback:
Great, maybe make command .time for user to show time?

SavvyNik:
I don't have think a command is needed. Later in newer clients the time is displayed on screen by opening the server name box in the top right corner.

GiladRe:
@janekback Sure, if you want it to display, just call in User.cpp CUser::SendTime(Paket &pkt) the g_pMain->SendHelpDescription function.

Client/N3Base/N3Camera.cpp
```diff
@@ -60,6 +60,7 @@ void CN3Camera::Release()
 	CN3Transform::Release();
 }
 
+// NOTE(Gilad): I don't think we use camera files, do we? as the load function never gets executed.
```
twostars: They will get used, ultimately. I know for a fact they're used by the 1.298 character selection screen, at the very least.

Server/shared/packets.h
```diff
@@ -146,6 +146,7 @@
 #define WIZ_PACKET17			0x8F
 #define WIZ_DEATH_LIST			0x90
 #define WIZ_CLANPOINTS_BATTLE	0x91 // not sure
+#define WIZ_FOG_DENSITY			0x92 // Controls the Fog density(Sorry for the random place. At some point i believe we'll need to re-organize the packets).
```
twostars: Not sure how I feel about custom packets. These overlap with official opcodes, so really they should be moved to some unused part of the range (note: the end is taken by special things). I think ~0xca to 0xfd is free.

Server/shared/packets.h
```diff
@@ -146,6 +146,7 @@
...
 
 // NOTE(srmeier): testing this debug string functionality
 #define WIZ_DEBUG_STRING_PACKET 0xFE
```
twostars: Sidenote: This one overlaps, as well (I wasn't including this in my previous comment).

Client/WarFare/GameProcMain.cpp
```diff
@@ -2269,6 +2288,13 @@ bool CGameProcMain::MsgRecv_Regen(Packet& pkt)
 	return true;
 }
 
+void CGameProcMain::MsgRecv_FogDensity(Packet& pkt)
+{
+	int16_t iValue = pkt.read<int16_t>();
+
+	s_pEng->CameraGetActive()->m_fFogDensity = iValue * 0.01f;
```
twostars: This really should be using a setter. :)

Client/WarFare/GameProcMain.cpp
```diff
@@ -5615,11 +5615,41 @@ void CGameProcMain::ParseChattingCommand(const std::string& szCmd)
...
+				iYear = stoi(sDate.substr(0, pos));
+				sDate.erase(0, pos + 1);
+				pos   = sDate.find(".");
+				iMonth= stoi(sDate.substr(0, pos));
+				sDate.erase(0, pos + 1);
+				iDay  = stoi(sDate);
+
+				if (iMonth < 0 || iMonth > 12) iMonth = 1;
+				if (iDay   < 0 || iDay   > 31) iDay   = 1; // TODO(Gilad): Was too lazy to implement the day for each month logic..please forgive me.
+			}
```
twostars: The range here is kind've screwy.
You're allowing 0 to 12 and 0 to 31?

I think the intent was either 0 to 11 or 1 to 12, but you get an extra here. Probably what you wanted was <= 0.

Server/GameServer/User.cpp
```diff
@@ -1180,6 +1180,37 @@ void CUser::SendTime()
...
+	int16_t hour  = pkt.read<int16_t>();
+	int16_t min   = pkt.read<int16_t>();
+
+	if (year  != NULL) g_pMain->m_sYear  = year;
+	if (month != NULL) g_pMain->m_sMonth = month;
+	if (day   != NULL) g_pMain->m_sDate  = day;
+	if (hour  != NULL) g_pMain->m_sHour  = hour;
+	if (min   != NULL) g_pMain->m_sMin   = min;
+
+	if (year == NULL && month == NULL && day == NULL && hour == NULL && min == NULL)
```
twostars: None of these are nullable types (because they're not pointers). The only reason this works is because NULL evaluates to 0, but this it not obvious.
This is why we use nullptr over NULL now; this would fail to compile there. NULL should only be used for pointers (and even then, we should be using C++11's nullptr in place of it).

Further, I don't think this is exactly what you wanted.  You want to set the time but only if it's not set to 0... but there are cases where this is intended, e.g. in your sample with time 18:00 -- the minutes portion is 0. Yet since that's 0, it would not update that so the minutes would not be updated.

Server/GameServer/User.cpp
```diff
@@ -1180,6 +1180,37 @@ void CUser::SendTime()
...
+	{
+		g_pMain->SendHelpDescription(this, "Using Sample: /time 18:00 2002.7.1 (Using the second parameter is not must(date format: Year.Month.Day).");
+		return;
+	}
+
+	Packet result(WIZ_TIME);
+	result << uint16_t(g_pMain->m_sYear) << uint16_t(g_pMain->m_sMonth) << uint16_t(g_pMain->m_sDate)
+		   << uint16_t(g_pMain->m_sHour) << uint16_t(g_pMain->m_sMin);
+	Send(&result);
```
twostars: I think we could probably just call SendTime() here instead of redoing this packet, but that's just a minor nitpick. :)

Having said that, should the time not update everyone?
In which case, you should be looping over everyone in-game and calling SendTime() for them.

Realistically since this updates server stuff, I think this should just be handled in CEbenezerDlg to begin with (or CGameServerDlg or whatever they renamed it to). Read out date/time, forward it to CEbenezerDlg to set and send to everyone.

Client/WarFare/PlayerMySelf.cpp
```diff
@@ -777,23 +777,30 @@ bool CPlayerMySelf::IsAttackableTarget(CPlayerBase* pTarget, bool bMesureAngle)
...
-	vDir *= mtxRot; // ȸ�� �� ����� ���ϰ�.
+	vDir *= mtxRot; // Find the direction of rotation.
 	if(fSpeed < 0)
 	{
 		fSpeed *= -1.0f;
 		vDir *= -1.0f;
 	}
 	__Vector3 vPosNext = vPos + (vDir * fSpeed); // ��� �ġ ���..
-	if ( false == ACT_WORLD->IsInTerrainWithTerrain(vPosNext.x, vPosNext.z, vPos) )
+	if (s_pPlayer->m_InfoBase.iAuthority != AUTHORITY_MANAGER &&
```
twostars: s_pPlayer is (the only instance of) CPlayerMySelf, which we're in right now (CPlayerMySelf::CheckCollision()). We can just access m_InfoBase directly :)

Client/WarFare/PlayerMySelf.cpp
```diff
@@ -829,7 +836,7 @@ bool CPlayerMySelf::CheckCollision()
 			if(fMag < fHeightSum) // �Ÿ��� Ű�� �պ��� �����..
 			{
 				float fMag2 = (pUPC->Position() - vPosNext).Magnitude(); // ����ġ�� �� ������.
-				if(fMag2 < fMag)
+				if(s_pPlayer->m_InfoBase.iAuthority != AUTHORITY_MANAGER && fMag2 < fMag)
```
twostars: In the case of the above check, this player and below NPC check, and the shape collision check, could we not simply wrap these all in one conditional to exclude them from being handled in their entirety as a GM?

Those don't set anything.
Edit: Minus the shape collision check, since we're using that to determine height level via incline.

Client/WarFare/GameProcMain.cpp
GiladRe: I guess i was a bit rushy on this check. Fixed.

Client/WarFare/PlayerMySelf.cpp
giladreich: Exactly what i though which is why i commented descriptive info on this function. But since it updates the next y coordinates while doing collision checks, i haven't really had the time to refactor this.

Server/GameServer/User.cpp
giladreich: Changed to == 0. 
Yea, that's right regarding the minutes being :00 in certain situations. I had this in mind, but since it will anyway read the packet as 0 when it's empty, there is no point updating stuff in the pointer for CGameServerDlg and for that purpose i included a check there. There is a different issue with the server not resetting the packets for some reason which at the second time you'll use the command, the server will read the old values. I wanted to create an issue for it, but i didn't really had the time to explore this more to add useful details.

Server/GameServer/User.cpp
giladreich: The reason for overloading this function is because CUser::GameStart(Packet & pkt) already using it.
I though it already updates the server time for all players? as g_pMain is a pointer to CGameServerDlg(if that's what you were referring to CEbenezerDlg?).

Client/WarFare/GameProcMain.cpp
giladreich: Many things here needs to get refactored with getters and setter. But as I'm not very familiar with the source code and doesn't have a big view of the whole project, I'm just fitting my self to the coding convention/style they used there, even though we know there is a lot to change and improve.
But anyways, added a setter :)

Server/shared/packets.h
giladreich: Changed packet to 0xca. Thanks for the heads up so far.

GiladRe:
Made some changes, I'll try to push the last commit improvements later tonight.

Server/GameServer/User.cpp
twostars: No, it updates the server time but it only immediately notifies the player who changed the time. You need to send this packet to all in-game players for the change to be made to those players.

SendTime() merely sends the packet containing the time info to the clients. This is why I said that instead of reimplementing this logic to send this (same) packet to the player, it would've been preferable to just have your method call the existing method to do so (no point duplicating logic, after all).

And yes, "CGameServerDlg" is CEbenezerDlg. Somewhere along the line someone renamed this.

Server/GameServer/User.cpp
twostars: The problem still stands, though. I'm well aware of how the packet handling works; I wrote it, after all.

The problem here is that you're having to "check" for this, and you're doing so in an unclear manner.

Firstly, why is this even an issue: why are we not providing all of this data at all times?
The official packet structure is as follows:
```
		uint16_t year, month, day, hour, minute;
		pkt >> year >> month >> day >> hour >> minute;
```
That's the structure. All of the data is always provided -- always.

If we're not always providing it in the client, the client is doing it wrong.
To address this, we can always send what the client deems the current time in place of blanks, and hence always send it a value.

Now if this wasn't an official packet (but it is, so we should treat it accordingly), we could otherwise address this in a couple of different ways.
We could:
1. design the packet such that this behaviour is obvious, and explicitly specify what we're changing or,
2. keeping it similar to the above, we could instead check to see if there's more to read from the packet.

Note that the existing checks make no sense, because if at any point something "isn't set", the rest "won't be set" either -- because there won't be anything left in the packet to read from.

The behaviour really doesn't make sense.

But to summarise, since this *is* an official packet, we should make sure that it behaves as it does officially. That is, the client will send _all_ of the date/time data, and the server will handle _all_ of them, always. If we exclude data from the client, it will still supply the current date/time data to the server.

Also note that this date/time isn't actually realtime. It moves at a speed much faster (I don't remember how fast, 6x? 10x? I'd have to check) than realtime, so we also don't want to be getting that mixed up when we do that. :)

Client/WarFare/GameProcMain.cpp
twostars: That is true, but we don't want to be adding to these (many) cases. Just removing them slowly :)

GiladRe:
Sorry for the delay, I'll probably have another look on the weekend as I'm currently renovating the flat.

Server/GameServer/User.cpp
giladreich: Oh...I think i get what you're saying. So the /time command doesn't really updates the server time? It's just sends a packet to the server and the server sends back to the client so the weather basically changes.
I'll try to look at it now and update the pull request.

GiladRe: I just realized that the CGameServerDlg::UpdateGameTime get executed on a background thread that gets the local time of the server machine every 6 seconds from DateTime wrapper class.
Which is why it got me confused and i implemented the /time command to parse also date(which is absolutely not necessary in this case).
But i commented it out in case someone will want in the future to parse date also for whatever reason and the server will support it if needed.
Will update the pull request soon.

GiladRe:
So this is the kind of behavior we want for /time command to do? now it updates for all online users:
capture

I completely forgot i can use the official binary client for debugging purposes. ^_^
@janekback I also added so it notify the user who executed it when the time is modified.

GiladRe:
Pull request #89 is now updated with the new changes.

Client/WarFare/GameProcMain.cpp
```diff
@@ -5756,38 +5758,57 @@ void CGameProcMain::ParseChattingCommand(const std::string& szCmd)
...
+				{
+					iHour = tm.tm_hour;
+					iMin  = tm.tm_min;
+					if (!sDate.empty())
+					{
+						iYear  = tm.tm_year + 1900;
+						iMonth = tm.tm_mon + 1;
+						iDay   = tm.tm_mday;
+
+						// Logic for max days in a month.
```
twostars, line 5798: All of the logic from here on is pointless; tm_mday -- and the other fields -- should always be valid when the structure's updated, which std::get_time() will be doing.

Server/GameServer/User.cpp
```diff
@@ -1197,40 +1199,64 @@ void CUser::UpdateFogDensity(Packet & pkt)
...
 {
 	if (!this->isGM()) return;
+ 
+	uint16_t iYear  = pkt.read<uint16_t>();
+	uint16_t iMonth = pkt.read<uint16_t>();
+	uint16_t iDay   = pkt.read<uint16_t>();
+	uint16_t iHour  = pkt.read<uint16_t>();
+	uint16_t iMin   = pkt.read<uint16_t>();
+
+	if (iHour == 0)
```
twostars, line 1225: This is going to break on midnight, no?
```
tm_hour | int | hours since midnight | 0-23
```



Again, we shouldn't be checking if these are 0 if what you really mean is they don't exist in the packet. But they're always intended to be in the packet, so there's no reason to check this regardless.

giladreich: That's correct. Then what you suggest? The work around i have in head atm as the packet is always exist, is just initializing(iHour, iMin...) all values with 99 as a dummy value for example and if the time in server is still 99, that means-> not update or failed to parse the time:
```
	if (iHour == 99) // 99 dummy value
	{
		g_pMain->SendHelpDescription(this, "Using Sample: /time 18:23");
		return;
	} 
```

It's ugly, but if you have a better suggestion, then let me know.
I do want to check there as i want to update the user when time mistype or modified.

Client/WarFare/GameProcMain.cpp
giladreich: It's not must to have all fields in tm structure completely for using get_time. The only reason i use the structure here so i don't need to write many different checks when it fails to parse the command string. That way the user won't crash the client when incorrect input received.

Server/GameServer/User.cpp
twostars: I think your problem here is trying to enforce your own logic/behaviour to an existing, official command/packet.

Rather than trying to workaround an issue created by trying to enforce this custom logic, I'd suggest going back and seeing how it works in the official 1.298 client, and then having it work the same here.

This is why I'm very much against custom behaviour. It's overcomplicating what should be trivial, and causing confusion where it doesn't need to exist.

twostars:
I manually merged in the changes that didn't include custom behaviour (aside from the sky fog behaviour, which doesn't affect anything of importance). 

There's a ton of work that needs to be done on this without having to maintain and allow for custom behaviour, so it's just best we just keep it in forks for now. Perhaps in future when the project is in a more stable place we can have a branch for it, but certainly not now.

Also, I'd have merged in those changes a long time ago if they were isolated and not in one giant pull request of miscellaneous things. Use branches for this, e.g.:
https://github.com/srmeier/KnightOnline/pull/91

Makes life so much easier and I don't feel nearly as bad for waiting on merging in things that are fine to do so (because they're mixed in with changes that aren't).

twostars:
#101 fixes all that was broken with the /TIME command, i.e. just the server not handling the packet. The client was already handling everything correctly.
This is why it's very important to see how things behave in the official client instead of assuming things work as you might expect them to :)

GiladRe:
To be honest, i definitely misunderstood how the /time command works. At the beginning i though it needs to update also the server time and should be parsed with the 2 dots (e.g /time 21:30) instead of spaces and this is why i also implemented parsing date because of the way the packet is structured in the server, which is why it over complicated things. 
I already knew that the server doesn't support it, which is why i removed it on commit ddf8bd1 the check you did for the packet opcode and wrapped it into a different function instead of using it in UpdateWeather, because in my opinion it's cleaner when each command has it's own implementation.
But since /time command doesn't updates the server time, it kind of make sense to use it in UpdateWeather even though it's not very obvious, because it's like you say here, "We updating time, but hey, we also updating weather, right? ;)", but when you see the packet structure on the server, you asking your self, "Hmmm....what does date has to do with weather??? xD", i hope you get what i mean lol. 

Regarding the /fog command, yea, maybe it's a bit custom, but it's more of a backend thing and it doesn't effect anyone. for that reason i included in my description that it's useful when debugging things such as rendering etc... But if you feel like merging it, let me know, because i stashed that one and can add it into a new pull request on a different branch if you like. Probably on the weekend I'll have time. 

P.S: Do you have any links to the official untouched clients of different KO versions? Because the only thing i found on my hdd atm, is an old client that is already modified. Looking for untouched files. So if you have versions:
10xx, 12xx, 15xx, 18xx, that would be cool. Because i remember it was possible to get it from kingofknight.com, but i couldn't find it unfortunately.

janekback:
http://www.kingofknight.com/patches/KnightOnlineSetup1453.exe
https://mega.nz/#!NEUXhCYS!um4-FFiIzCs8S9MJ-HWPfZ4L26NO1EWgFaNPTYg8t_w 
The second link is 1299 client.

  have nice day :)

GiladRe:
Cool! Thank you @janekback! Have a nice day too! :) 
If anyone know where can we find the 10xx and other client versions, that would be also nice to have.

PR #90, opened 2017-09-10
Implement Registration/Homepage-Button on Loginscreen

madpew:
Implemented the Registration-button on the Loginscreen by reading the website-address from the Server.ini (as the 1299 binary client does) and opening it using shellexecute.


twostars:
I think these UIs should really be handled separately; these buttons don't exist in the 1.098 UI (unless I'm crazy). Even the indexes for these UIs in the TBL are different (they both exist in the client), so I'd be inclined to leave these for another implementation of the class specifically for the 1.298 UI.

twostars:
Nevermind, I'm crazy. Not sure how I've never noticed that button there this entire time. :)

PR #105, opened 2018-02-28
Added WIZ_QUEST packet handling to client side. It give information i…

d5dl2:
…f user start a quest. It's still need development.


srmeier:
Looks good to me.

PR #106, opened 2018-02-28
I am trying to add change page using mouse wheel function to Skill Tr…

d5dl2:
…ee Page (K), Transaction Npc Page's and Inventory Page's. I dont know if this is correct way but its work fine.


d5dl2:
This request has incomplete changes. I completed it on #114 , so we can close this.

PR #107, opened 2018-02-28
Coins showing at inventory dialog seperated by comma. The codes that …

d5dl2:
…do the same job are collected into method.


srmeier:
Looks good, I suppose for CUIInventory::GoldUpdate() that's new functionality for the CN3UIString "text_gold"? I guess the difference is that here you're adding the commas in the right place?

d5dl2:
Yep, i added commas only "text_gold" named CN3UIString. It is showing like "xx,xxx,xxx". But for now it works only for inventory view. It needs to be implemented other views like transaction, warehouse etc.

Issue #104, opened 2018-02-28
Errors on loading some fx files


d5dl2:
There are access violation errors while reading fx binaries. When player use some lightning skills to monsters or when you hit the monster with +8 stiletto, access violation error occurs on 
 **_bool CN3FXPartBase::Load(HANDLE hFile)_** method in 
 **.\Client\N3Base\N3FXPartBase.cpp** file. I encounter this error using both two data.zip that you share before.

`m_ppRefTex = new CN3Texture* [m_iNumTex];` on this line of code, `m_iNumTex` has very big value something over millions and program cant allocate that big memory.

When i debug; it also cant load `m_iVersion` from file at start of method. 
We may prevent this issue by adding 
`if(m_iVersion == 0) return false;` line, it works nice, game can be continue to play, complete or less effect happens in the screen (i dont know effect really completed on screen). But it is not right way to fix it i think. What i want to understand is:

- Why code still cant load properly data files although you have been attached to support different versions of files.
- How i ll learn to how fx files can be read properly. I mean how i may know structure of data files.
- How i may fix this kind of errors, i can't see the start point. If i change the function, it may be broken reading other files. I dont know that file structure and how it need to be.
- If you want to just support 1298/9 client, why you want to support also older data files instead of just force update the user who using older files ? 

_At the and: sorry for my bad english._

srmeier:
This is likely due to a variation in the FX file format. Reading in incorrect binary information from the file, or uninitialized variables, often result in these very big (unrealistic) values.

srmeier:
Do you happen to know the name of the lightning FX file that causes this issue? I can load this file manually and see what's happening here.

koraykarakus:
this is because of 2bytes of data forgotten to be read, so after first read we start reading from wrong indice.

see N3FXBundleGame.cpp

inside --> bool CN3FXBundleGame::Load(HANDLE hFile)

replace if block ---> if(m_iVersion>=1)

with the version below, it basically SetFilePointer to start reading from correct place ( 2bytes to the right ), and remove that modification on continue, because we did not read. All healing skills, buff, ac, fresh mind, rogue evade/sacrifice, mage skills are working fine with this modification. But better understanding about that 2 byte is required.

`if(m_iVersion>=1)
{
	for(int i=0;i<MAX_FX_PART;i++)
	{
		int iType;

		if (i != 0)
		{
			//push right 2 bytes, need to read 2 bytes of data here -- 01 01 hex
[...]

PR #108, opened 2018-02-28
to prevent access violation errors while reading FX files.

d5dl2:


srmeier:
@twostars Do you know off the top of your head whether we should have a check for client version here?

onurcanbektas:
@d5dl2  Have you checked where $m_iVersion buffered and why is not buffers in your case ?

twostars:
I don't think this is actually fixing the root problem here, I think it's just masking it.

Can you give us a sample of what's causing these issues?

d5dl2:
I was try to give sample about this on #104 
This issue happens when i use some lightning skills to mobs on mage class or attacking mobs with +8 stiletto (poisonal) on rogue class. I aggree with you it is just masking issue, it may cause playing the game without unexpected disconnects until we fix main problem.

giladreich:
```
// In between versions max part count. Version 3 and above we can use up to 26 parts in one bundle.
const int MAX_FX_PART_V0 = 8;  // V0
const int MAX_FX_PART_V1 = 16; // V1 till V2
const int MAX_FX_PART_V3 = 26; // V3 till V6
//const int MAX_FX_PART_V7 = 30; // Who knows, just a guess! :P Looking forward to see versions 7 ^_^

const int MAX_FX_PART = MAX_FX_PART_V3;
```

@d5dl2 You probably want to look at  bundle instead of the part. Hope that makes sense :)

srmeier:
Do you happen to know the name of the lightning FX file that causes this issue? I can load this file manually and see what's happening here.

PR #109, opened 2018-02-28
Handling Mouse Middle Button pressed event. It need some work to turn…

d5dl2:
… camera 180 degree.


srmeier:
"CameraYawAdd(100)", I would naively expect this to be 180. Also, is the rotation instantaneous or is it animated? Might need to test this further against the 1299 client.

d5dl2:
It is not turning camera exactly 180 degree neither yaw +100 nor yaw +180. This implementeion turns the camera instantaneous. This need some work to be like offical 1299 (Animated 180 degree turning)

PR #110, opened 2018-02-28
Error on using skill points to assasin on rogue and passion on warrior.

d5dl2:
Using skill points to assasin on rogue and passion on warrior

### DIFF
--- a/Client/WarFare/UISkillTreeDlg.cpp
+++ b/Client/WarFare/UISkillTreeDlg.cpp
@@ -573,7 +573,8 @@ void CUISkillTreeDlg::PointPushUpButton(int iValue)
 				break;
 		}
 	}
-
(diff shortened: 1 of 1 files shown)


srmeier:
Can you explain this one a little more? What was the error (perhaps link the issue number) and why is commenting out this piece of code a good fix?

d5dl2:
On Human or Karus Rogue, when you have skill points to use, you cant use them for Assasin Tab Skills which iValue = 6 referenced. This means, you cant even have 1 skill point on assasin. 

Same goes for Warrior (Human or Karus), when you have skill points to use, you cant use them for Passion Tab Skills which iValue = 7 referenced. This means, you cant event have 1 skill point on passion.

I didnt understand what is the purpose of this code.

srmeier:
Ok I can see where you're coming from. For now we can keep it commented out, especially if it doesn't serve an obvious purpose, and perhaps we can consider removing it once the skill system is fully fleshed out.

PR #113, opened 2018-02-28
Pull request#106 configurations

d5dl2:
I m new on using Git and SourceTree. I was miss to add a file and changing on configuration files to my previos branch and i was allready created pull request for it. This request is completion of request #106


d5dl2:
I was make a mistake, im gonna create new request with fixing.

Issue #40, opened 2017-04-04
Navigating with mouse wheel in Skill Tree


Demircivi:
You were able to switch between pages by using mouse wheel in Skill Tree in 1299.

twostars:
This issue was moved to ApexKO/OpenKO#40

d5dl2:
This problem handled by #114. Also #41 and #42 handled too. We may close them.

PR #114, opened 2018-02-28
For Issue #40#41#42 - PullRequest#106 Completion

d5dl2:
I m new on using Git and SourceTree. I was miss to add a file and changing on configuration files to my previos branch and i was allready created pull request for it. This request is completion of request #106 

I may have mixed things together :D This request just has additional updates commit (2d4b542) for f41ce82 commit.

### DIFF
--- a/Client/WarFare/IMouseWheelInputDlg.h
+++ b/Client/WarFare/IMouseWheelInputDlg.h
@@ -1,0 +1,7 @@
+#pragma once
+class IMouseWheelInputDlg
+{
+public:
+	virtual bool				OnKeyPress(int iKey) = 0;
+};
+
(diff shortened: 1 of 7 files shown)


srmeier:
Can you add a quick comment about the code here? If I'm understanding correctly you are adding this to treat pageup and pagedown differently on the UI skill pages? Is that right?

d5dl2:
Yes you are right. I create a new interface "IMouseWheelInputDlg" which has OnKeyPress event to bind Mouse Wheel input for UI change page functionality. 

On LocalInput.cpp, i have handled mouse wheel inputs for only UI's which inherited from IMouseWheelInputDlg interface, then use its function for request page up/down events. 

In the future, if we need to implement same (Mouse Wheel Input) to another UI, just we need inherit that UI from IMouseWheelInputDlg and handle OnKeyPress method for UI, i think it will work.

But i dont know, where these codes (that i writed in LocalInput.cpp) should be.. It may handled by existing or new function anywhere else which may more related about UI.

srmeier:
Ok that makes sense for now. Thank you for all your help d5dl2!

PR #116, opened 2018-03-01
Changed TOGGLE_HELP function from F1 to F10

d5dl2:
Handled F1:F8 keys to meaningfull enums
Changed TEST_PACKET function on debug mod from F8 to F11 (which is available and sensible key for now)
Handled F1:F8 key presses to change hot key dialog page.


srmeier:
Are the F key assignments here the same as the official client?

d5dl2:
Yep i think it is same as offical client which F1-F2-F3...F8 connected to HotKeyDialog Pages, F10 connected to Help Dialog.

PR #117, opened 2018-03-01
Skill icons which not available

d5dl2:
About Skill Icons which are not available
In skill tree dialog, there are no icons for skills which not available on our level
I try to implement it, also i create new function for checking the skill if its available on our current level, when fires right click on icon (for put it hot key dialog).
I dont know this is the right and easiest way.


srmeier:
Looks good!

PR #120, opened 2018-07-19
Lua Bindings and 21.lua corrections

mjsamet:
Added Lua Bindings: **CheckExistEvent, CheckNoExistEvent** for quests
**EventID  21130** corrections

Added CheckExistEvent, CheckNoExistEvent, ZoneChange to **Evt Converter**


srmeier:
Looks good to me! Looking over the code I believe you also addressed @twostars comments made in PR #119. Will merge this, thank you for your hard work.

onurcanbektas:
Does any one knows why we are still keeping lua files in the repo ?

srmeier:
> Does any one knows why we are still keeping lua files in the repo ?

Good question, they are required for questing but if people are able to generate them automatically from the EVTs then perhaps we can remove them from the repo.

onurcanbektas:
Then perhaps we should keep them. Beginners might have some problems about generating them.

tra666:
Nice

Issue #124, opened 2018-09-01
Knights chat button is broken


ohad258:
Knights chat button is broken, since the client code is not compatible with the 1299 data files.
Small fix is required

PR #125, opened 2018-09-01
KO-124 fixed knights chat button name

ohad258:
#124 fix,


srmeier:
Hi @ohad258, sorry for the delay on this! Looks good, thanks!

I could perhaps see this being an issue if we wanted to support the older UIs, as I assume maybe it was originally referred using `btn_clan` but that ultimately changed for 1298. Not sure though!

Merging.

ohad258:
Hey @srmeier 
So we clearly want to be compatible with the 1299 client files.
But what about incompatibility between the server and the client?
I found a mismatch with a particular packet parsing between the Game Server and the Client, and wondering where I should apply the fix.
Any thoughts? :)

srmeier:
@ohad258, The packet structure in the client is likely the older 1098 packet structure. The game server packet structure is likely 15XX (in places where it doesn't match the client). In order to identify the true 1299 packet structure we would use the official leaked server binaries (if you don't have them reach out to me on Skype) and we would use a program called SocketSniff (there are others that will do it too) in order to identify what the actual packet should look like.

PR #146, opened 2024-10-10
Strip SDL 2

twostars:
SDL 2 was sort've hamfisted in to try and encourage support of other platforms, however this is very much coupled to DirectX and other Windows functionality, so all this has really lead to was:
1. Hacking around to try and make things work with it,
2. Not very ideal resource usage (having to load things like cursors from external assets, meaning more assets to include),
3. Awkward dependencies to software that is already long outdated and would have to be rewritten.

I think it's preferable to just focus on development for Windows first, and then plan out a more reliable approach to handling other platforms later (ideally such a thing would've been done before development of the project itself, but that wasn't really in our hands!).

Keeping things native to Windows also means it's easier to verify behavioural accuracy, without running into subtle issues requiring delving into the internals of SDL2 to figure out.


twostars:
Given the activity of the project, I'll just merge this one in now.

PR #156, opened 2025-05-01
Tbl editor for OpenKO

koraykarakus:
In order to address the language-related display issues in TBL files used by OpenKO, a simple graphical TBL Editor was developed using MFC (Visual Studio 2022).

Key Features:
Allows basic viewing and editing of .tbl files

Integrated with the existing resource structure of the project

Designed for ease of use and further expansion

This tool aims to make TBL file modifications more accessible for developers working on OpenKO or similar projects.


twostars:
This isn't really ready to merge in. It didn't even compile originally due to the inclusion of Turkish, since it's compiling using the system's codepage and not being told to treat it as UTF-8 or whatnot.

No real reason to use Turkish here anyway; it's an English project.

I went through and started cleaning it up a bit, but I'm very tired so I'll go back over it at another time.

The CString/std::string stuff was a bit of an unnecessary problem depending on which build config was used, since it was assumed to be ANSI only for conversion but other build configs are Unicode. Wasn't really a reason to keep it using std::string anyway; the only time we need to read them a certain way is from the TBL. We should *ideally* attempt to decode it appropriately/contextually, but for now I've just left it placeholdered with the system codepage, which is just how it's handled by the client anyway.

The UI is also not very functional. It kind've works; it's OK for reading, but modifying is very painful.
Even just inserting a new row defaults columns to empty strings instead of using their column-appropriate defaults (e.g. "0.0").

I think moving the edit control over the top of the field we're modifying and editing that is a start; it's very clunky as it is.
Strings could probably be handled specially for better multi-line support.

Also: when making a pull request, it's important to only include the changes relevant to that PR.
Each of your PRs cumulatively added each change so I had to go through and rebase each of them to strip them (i.e. `git rebase -i [our_master]`). If you could avoid doing that in future, it would save the hassle.

e.g.:
image

I think your problem is that while you have each in their own branch (so you're halfway there), the base for each branch is that of your master branch at the time of adding the change, NOT *our* master branch.

It's maybe a little complicated to explain, but you might consider doing something like:
1. Create either a dev branch or a branch synced with our master branch (upstream/master) so there's some degree of separation.
2. Whenever you make a change, swap to the branch synced with our master branch (upstream/master) and then create a new feature branch off that (e.g. `git checkout -B fixed-a-bug`).
3. Now when you create a pull request for this, it only contains your feature branch changes -- nothing else. We won't have to untangle it.
4. If you're wanting to use those changes before we merge in the PR in this repository, you can swap back to your dev branch, merge in the changes from your feature branch (`git checkout dev`, `git merge fixed-a-bug`) and just carry on with your day.
5. When we *do* update, you can basically either just merge it in (creating a bit of a deviation), or alternatively rebase your dev branch and basically strip out the original/early additions. Just merging it in is nice and safe though, so I'd recommend that.

I'm not really an expert on this so I don't know if there's a better/nicer way to go about it, but that's essentially what I do.

PR #162, opened 2025-05-03
fix for skill drag drop from skill tree to hotkey bar

koraykarakus:
Probably for testing purposes some part of the code is commented out and that causes drag / drop feature to not work from skilltree to hotkey. Re adding those solved the presenting issue.


twostars:
Looks like it was done when bitshift initially went over the codebase and updated it from VC++6. `i` would have been accessible below, so it was probably just commented out at this point without too much extra thought.

PR #163, opened 2025-05-03
attempt to fix the issue all skills share the same cooldown

koraykarakus:
different skills share the same cooldown because of commented outed part of the code. This need further investigation and possible new bug fixes. However, at the current state game is almost not playable for most classes like warrior, rogue and priest. Some single line if statements also converted because of testing difficulties encountered with TRACE macro. Used Trace lines removed out of code, and afx.h also removed but single line if statements converted to multilines for simplicity in further attempts. 

In short, please take a look at line 1116 for the important change which fixes the issue.

### DIFF
--- a/Client/WarFare/MagicSkillMng.cpp
+++ b/Client/WarFare/MagicSkillMng.cpp
@@ -21,7 +21,6 @@
 #include "N3SndObj.h"
 #include "N3SndObjStream.h"
 #include "N3ShapeExtra.h"
-
 //#include "StdAfxBase.h"
 
 #ifdef _DEBUG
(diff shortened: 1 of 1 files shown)


twostars:
Officially this is handled by the placeholder fields in the `__TABLE_UPC_SKILL` and managed cooldown lists.
ProcessCombo() is also useless; this was officially removed.

I won't be merging this in, in this state as it doesn't really address the issue. All it does is it removes the enforced cooldown outright, which is not what we want at all.

twostars:
Closing as #169 attempts to replace the functionality instead of outright removing it

PR #166, opened 2025-05-06
NTF7 Encryption support

UTengine:
To be honest I really did my best here I'm not a programmer and I don't know much what I'm doing but I always try to give my best shot.


UTengine:
Thanks for cleaning up, creating a nifty wrapper and merging it <3.

PR #169, opened 2025-05-08
Skill recast map fix

UTengine:
Simply put it works, I haven't tested potions items only mage skills but in theory they should work.
Not sure what to do with legacy code as it confuses me but this is better than what we have.
It does the job the only issue I came across is you cant Z (select nearest enemy) or select a monster with your mouse for 1-2 seconds after casting ends not sure if its related  to the packets we are sending the same .cpp or elsewhere like gameproc.

Addresses the following PR/issue: https://github.com/srmeier/KnightOnline/pull/163

image

### DIFF
--- a/Client/WarFare/MagicSkillMng.cpp
+++ b/Client/WarFare/MagicSkillMng.cpp
@@ -1126,7 +1126,7 @@ bool CMagicSkillMng::MsgSend_MagicProcess(int iTargetID, __TABLE_UPC_SKILL* pSki
 	auto itRecast = m_RecastTimes.find(pSkill->dwID);
 	auto itNonAction = m_NonActionRecastTimes.find(pSkill->dwID);
 
-	if (pSkill->iSelfAnimID1 >= 0) // Casting skill with animation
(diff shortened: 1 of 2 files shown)


UTengine:
maybe we should re-tie the rest of the values >.>

twostars:
I would remove them when they expire just to keep the maps small. They're not bound to grow that big anyway, but every little bit helps.
Less to iterate over, faster to process, less memory involved.

UTengine:
Ok ill check it tomorrow and when while at it I will also put a breakpoint on Z and some other regions try to debug it.

UTengine:
Not sure but I think the delay bug has to do with iSelfAnimID ? like when you step backwards u can cast the next skill repeatedly if you dont you have to wait for quite a long delay.

UTengine:
Ok obviously its returning iscasting true when it shouldn't because the cast animation was over already so I have no Idea why the game behaves like this.
Nvm something doesnt seem to be resetting the players state or something I'm really lost to this behavior.
	if (s_pPlayer->State() == PSA_SPELLMAGIC ||

PR #170, opened 2025-05-09
Stat skill reset

koraykarakus:
I was working on this before new update of GetText() so my LoadStringFromResource should be replaced with new method. This is a working skill & stat reset for openko. We can use it for testing. Some string could not be found in .tbl files ( Texts_us ) so new not original messages generated. Maybe in the future text issues can be solved. Some client side controlls skipped ( like : isDead, isTrading etc. ) they can be added later, they have been checked on gameserver for now. 

For this to work
Server/bin/Debug/Quests/21.lua
Server/bin/Release/Quests/21.lua
where nEventID == 21951
change as 
elseif nEventID == 21951 then
	pUser:OpenStatSkillReset();
	do return; end
 should be changed as mentioned on top, I force pushed those because git don't track them. 
 
 Thank you for maintaining this nice project. Soon I will update latest version to commit properly.

#changed files

Client
GameProcMain.cpp
GameProcMain.h
UIMessageBox.cpp
UIStatSkillReset.cpp
UIStatSkillReset.h
GameDef.h ( button behaviors )

Server/shared
packets.h

Server
LuaEngine.cpp
lua_bindings.cpp
lua_bindings.h
NPCHandler.cpp
User.h
User.cpp

Server/bin/Debug/Quests/21.lua
Server/bin/Release/Quests/21.lua
where nEventID == 21951
change as 
elseif nEventID == 21951 then
	pUser:OpenStatSkillReset();
	do return; end

### DIFF
--- a/Client/WarFare/GameDef.h
+++ b/Client/WarFare/GameDef.h
@@ -1242,6 +1242,8 @@ enum e_Behavior {	BEHAVIOR_NOTHING = 0,
 					BEHAVIOR_PARTY_BBS_REGISTER_CANCEL, // 파티 게시판에 등록 해제
 
 					BEHAVIOR_EXECUTE_OPTION,			// 게임 종료하고 옵션 실행..
+					BEHAVIOR_STAT_RESET,				//user clicked confirm on stat reset
+					BEHAVIOR_SKILL_RESET,				//user clicked confirm on skill reset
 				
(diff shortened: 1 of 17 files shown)


twostars:
Already exists, see my comment here:
https://github.com/srmeier/KnightOnline/pull/195#issuecomment-2907535366

PR #171, opened 2025-05-10
Fix normal user sees gm command tab

koraykarakus:
Normal users were able to see GM command tab. It is fixed with these changes.


twostars:
Closed because of #174

PR #173, opened 2025-05-10
give_exp command for admin added, this can both add and remove exp fr…

koraykarakus:
a new command to fasten testing. Adding exp from database each time and dealing with sql queries and management studio is not effective.

instead of this now we can use /give_exp UserName 5000000 1
the last entry is to make Notice visible or not.

also supporting negative amounts like /give_exp Username -5000000 1

these can be used on nonGM users by game master.


koraykarakus:
this can be removed better version inside gm commands, this is buggy and don't consider min and max levels, so will result in disconnect at login screen.

PR #174, opened 2025-05-10
Hide gm tab for normal users no conflicts

koraykarakus:
This basically hides GM tab for non authorized users on UI. This PR is resubmitted because of conflicts.


koraykarakus:
that is not official behavior, normal user can see GM tab but they cannot see King tab.

PR #175, opened 2025-05-17
gm commands for game

koraykarakus:
Patch for admin commands which can be really helpful via testing the game at the current state. This commit includes:

1- give_item UserName ItemID [optional 1/0 for notice]
basically adds an item to user, it can be either GM or normal one. Code is expanded to support giving item with item name but it is unfinished because of lack of mechanism to search an item with its name. Maybe in the future CSTLMap can be expanded to support GetItemByName in addition to GetItem(ID)

2- give_exp Username ExpAmount [optional 1/0 for notice]
this can both add and remove exp from user, better than last comitted version. Now user cannot go beyond max level or cannot go down min level. Multiple level jumps handled with for loop so user now don't need to kill additional mobs after using the command.

3- arrestuser Username
GM teleports to the user

4- summonuser Username
GM summons the user

5- cutoff Username
GM disconnects the user

6- forbidconnection Username [optional 1 or 0 for notice]
notice option added

7- permitconnection Username [optional 1 or 0 for notice]
notice option added

8- forbidchat Username [optional 1 or 0 for notice]
notice option

9- permitchat Username [optional 1 or 0 for notice]
notice option

10- disableattack UserName [optional 1 or 0 for notice]
notice option.

11- attakcenable UserName [optional 1 or 0 for notice]
notice option. This also added wrong in TBL file Texts_us.tbl, that is written as attakcenable instead of attackenable
need fixing

12- zonechange UserName zoneID
teleport to target zone with zoneID

13- give_coin UserName Amount [optional 1 or 0 for notice]
give coin to user, it can be both positive and negative amount

14- visible/invisible also carried into admincommandhandler.cpp

15- gm_help 
this outputs gm commands ingame

16- removed WIZ_OPERATOR and switch block to prevent confusition

17- removed Admin_MsgSend and seperated it with different packets for each command.

### DIFF
--- a/Client/WarFare/GameProcMain.cpp
+++ b/Client/WarFare/GameProcMain.cpp
@@ -100,7 +100,8 @@ enum e_ChatCmd
 	CMD_VISIBLE, CMD_INVISIBLE, CMD_CLEAN, CMD_RAINING, CMD_SNOWING, CMD_TIME, CMD_CU_COUNT,
 	CMD_NOTICE, CMD_ARREST, CMD_FORBIDCONNECT, CMD_FORBIDCHAT, CMD_PERMITCHAT, CMD_NOTICEALL,
 	CMD_CUTOFF, CMD_VIEW, CMD_UNVIEW, CMD_FORBIDUSER, CMD_SUMMONUSER,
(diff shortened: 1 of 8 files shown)


twostars:
This was written for the old snoxd-koserver and is entirely unofficial behaviour, so I'll close this now.

PR #177, opened 2025-05-20
fx particle crash fix

koraykarakus:
this address issue #104 

fx files read wrong after 1st. Some data which is 2 byte long is not read at all and so the following readings are wrong and this causes game crash because of too big trash data from the file. 

2 byte offset has been add to fix this issue and to avoid crash simple controll has been added ( which is not be needed as long as file read correctly, but at this stage I am not sure about all the files / skills ) 

tested most of the skills works OK, some skill animations are incomplete like safety (rogue), and some are different from the original version ( like 45 mage skills, staff skills etc. )

### DIFF
--- a/Client/N3Base/N3FXBundleGame.cpp
+++ b/Client/N3Base/N3FXBundleGame.cpp
@@ -558,9 +558,19 @@ bool CN3FXBundleGame::Load(HANDLE hFile)
 		{
 			int iType;
 
+			if (i != 0)
+			{
+				//push right 2 bytes, need to read 2 bytes of data here -- 01 01 hex
+				SetFilePointer(hFile, 2, NULL, FILE_CURRENT);
+			}
+
 			ReadFile(hFile, &iType, sizeof(int), &dwRWC, NULL);
 
-			if(iType == FX_PART_TYPE_NONE) continue;
+			if (iType == FX_PART_TYPE_NONE)
+			{
+				SetFilePointer(hFile, -2, NULL, FILE_CURRENT);
+				continue;
+			} 
 
 			else if(iType == FX_PART_TYPE_PARTICLE)
 			{
--- a/Client/N3Base/N3FXPartBase.cpp
+++ b/Client/N3Base/N3FXPartBase.cpp
@@ -493,6 +493,10 @@ bool CN3FXPartBase::Load(HANDLE hFile)
 		else m_bAlpha = FALSE;		
 	}
 	
+	//this is not required as long as file read correct
+	if (m_iNumTex >= 19)
+		m_iNumTex = 1;
+
 	m_ppRefTex = new CN3Texture* [m_iNumTex];
 
 	std::string FileName;


twostars:
While this may stop it crashing, all you're doing is breaking the data of every effect with more than 1 part by doing this. Typically (though not exclusively), file format changes here are versioned, and typically (but again, not exclusively) those version changes are added to the end of each part (or bundle), so they're handled sequentially.

You're blindly skipping 2 bytes at the start of every part here aside from the first, while trying to fix a crash relating to the particle type only.
This just blindly breaks all the data for every part read beyond the first one.

Rather than try to hide a crash and introduce random other effect bugs as a result, it's better to just have it obviously crash so that it can be immediately identified and fixed appropriately.

koraykarakus:
at curret state data is already broken, path names not read correctly for example:

1st read : fx\particle\30304evade\evade
2nd read : 304evade\evade
14 byte missed because we did not skip that 2 byte and started to ready somewhere wrong. Because we read wrong places in file now our 

for example for evade (rogue skill) we need to start reading particles from 07 02 hexes each time to get correct but because we did not read a required data of 2bytes inside for loop now we are reading the first encounter of 01 00 00 00 and our next readings are all messed up.

there is something forgetten inside that for loop

example from evade.fxb

01 01 01 00 00 00 CD CC CC 3E 07 02 00 00 40 40 01 00 00 00 00 00 00 00 00 00 00 00 00 00 00 00 00 00 00 00 00 00 00 00 00 00 00 00 00 00 00 00 00 00 00 00 00 00 00 00 00 00 00 00 80 3F 00 00 00 00 01 00 00 00 00 00 80 3F 66 78 5C 70 61 72 74 69 63 6C 65 5C 33 30 33 30 34 65 76 61 64 65 5C 65 76 61 64 65

first 2 bytes not read ( assuming whoever write that for loop forgot to read them) so if we dont skip those 2 bytes iType is wrong

1st reading : 01 01 01 00 ( irrelevant number for loop does nothing )
2nd reading : 00 00 CD CC (another irrelevant )
3rd reading : CC 3E 07 02 
4th reading :  00 00 40 40 
5th reading :  01 00 00 00 ( now we got a good value which we seek for 1-2-3-4 one of them or 0 for continue )
and because of that 2 bytes we got huge number of iType + wrong path 

if we skip 2 bytes each time after 1st reading
01 01 skipped.
01 00 00 00 = 1 good number iType = 1 
CD CC CC 3E = good number float 0.4 (fStartTime)
07 - first reading m_iVersion inside load
02 - second reading inside load m_iBaseVersion 
00 00 40 40 - 3sn float m_fLife
01 - m_iType = 1
3 x 12 bytes of logical vector values ( otherwise they are huge numbers )
m_bOnGround = 0 
another 12bytes of vector
01 00 00 00--- m_iNumTex = 1 ( as expected ) 
m_fTexFPS = 1 (float) 
m_pTexName = 66 78 5C 70 61 72 74 69 63 6C 65 5C 33 30 33 30 34 65 76 61 64 65 5C 65 76 61 64 65 = fx\particle\30304evade\evade

this works for all skills (0-60 ) did not test others, but when I trace paths all the paths are correct and all the time `int iType;` is logical. 

I know it is hacky to fix it like as follows but what I mean either way we need 2 bytes of reading there

if (i != 0)
{
	//push right 2 bytes, need to read 2 bytes of data here -- 01 01 hex
	SetFilePointer(hFile, 2, NULL, FILE_CURRENT);
}

So what can we really read there ? which corresponds to hex "01 01". I assume uint8_t readings which follow each others maybe some version info ?

twostars:
The loop is fine. The problem is in reading the parts, as confirmed by both your title, and your code only skipping those bytes for every part AFTER the first part.
If it was a problem with the loop, you wouldn't have skipped the first part, because it would apply to every iteration in the loop.

So clearly the extra bytes belong to the previous part.
And since as you state, your problem is with particles, clearly the extra bytes belong specifically to the previous particle part.

Which means every effect that doesn't have only particle parts, you're essentially breaking all of their data (and apparently no longer directly crashing). Which is far worse than it simply crashing from not reading correctly in the first place and directly crashing.

PR #206, opened 2025-05-27
Magic skill mng fix

koraykarakus:
## Pull request type

- [x] Bugfix
- [ ] Feature
- [ ] Build related changes

At current state, warriors cannot cast skills with 2h weapons
Priests can cast attack skills with daggers & staff
Rogues can cast attack skills with 1hswords / staves / 1haxes ...
Warriors can cast skills with daggers & staff ( basic ones)
Mage can cast staff skills 42/72 with any weapons
Skills requiring items ( like using transformation gems which requires totems ) handled wrong

## What is the new behaviour?

Now priest can cast attack skills with all kind of weapons except for staff & daggers,
Priest can use buff skills with any weapon
Warriors can cast attack skills with all kind of weapons except for staff & daggers
Rogue ( assasin ) is limited to using daggers, they cannot use swords or axes to cast skills
Rogue ( archer ) cannot use any other weapon than bows for attack skills
Mage cannot cast staff skills with any weapon, only staffs.

dwNeedItem logic was wrong, it was just considering warrior skills ( 1055 & 1056 ). Also it was making impossible to use some
items like transformation gems, catapult etc.

dwNeedItem logic for casting skill used properly in switch block to have sustainable code for future additions ( like catapult items, transformation totem etc. ) 
## Checklist

### DIFF
--- a/Client/WarFare/MagicSkillMng.cpp
+++ b/Client/WarFare/MagicSkillMng.cpp
@@ -197,49 +197,164 @@ bool CMagicSkillMng::CheckValidSkillMagic(__TABLE_UPC_SKILL* pSkill)
 	if(pInfoExt->iMSP < pSkill->iExhaustMSP)
 		return false;
 
-	int LeftItem = s_pPlayer->ItemClass_LeftHand();
-	int RightItem = s_pPlayer->ItemClass_RightHand();
-
-	if(pSkill->iNeedSkill==1055 || pSkill->iNeedSkill==2055)
-	{
-		if((LeftItem != ITEM_CLASS_SWORD && LeftItem != ITEM_CLASS_AXE && LeftItem != ITEM_CLASS_MACE ) ||
-			(RightItem != ITEM_CLASS_SWORD && RightItem != ITEM_CLASS_AXE && RightItem != ITEM_CLASS_MACE) )
-		{
-			return false;
-		}
-	}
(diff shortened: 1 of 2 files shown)


twostars:
This seems vastly overcomplicated. Look at the numbers (dwNeedItem), and then look at `e_ItemClass`.

twostars:
The old logic is pretty much the original leaked logic. It had similar behaviour in the server; nobody's really touched it.
It's just very outdated and needed to be updated.

> dwNeedItem is skill's item requirement, if it was only designed for usable items like transformation scroll, catapults etc. It would not include class skills ( warrior skills = 0, priest skills = 0, assasin = 1, archery = 7, mage staff = 11 )
I understand that. And I'm also telling you to compare it to e_ItemClass; assassin (daggers), archery (bows), etc.
You should notice a pattern which should heavily simplify your logic there.

UTengine:
I had pushed this lately to ko4life and I believe it was hoe the original src should have handled it but just without the silly warrior checks and bugs.

twostars:
I'm going to close this one. This is very old and the server changes are no longer compatible.
Beyond that, this is definitely not the right way to approach this issue.

UTengine:
Just wanted to say afaik dwneeditem 9 is any weapon for melee skills afaik. And none for magic casting afaik. But I could be wrong, because in the texts.tbl any weapon/items was added and it did not exist there before.

PR #207, opened 2025-05-28
usable items to scrollbar from inventory drag & drop

koraykarakus:
## Pull request type

- [x] Bugfix
- [ ] Feature
- [ ] Build related changes

Transformation scroll and other many items ( such as delos war items ) cannot be put on skill bar from inventory.

## What is the new behaviour?

they can be drag & dropped now

UIITEM_TYPE_SONGPYUN_ID_MIN value set wrong as 490000 
class skills start from lower numbers and end around 300000, UIITEM_TYPE_SONGPYUN_ID_MIN is used to filter class skills but it has been started from 490000 instead of 450000.

450000 is the point where usable items start.

## Checklist


koraykarakus:
bpoint

Issue #192, opened 2025-05-22
no warning message when user tried to exchange abyss gem while inventory is full


koraykarakus:
need to know what kind of message output is expected ? Was it alertbox or output message on info chat.

koraykarakus:
this is not official behavior, officially no message is displayed even if invenyory is full.

PR #208, opened 2025-05-28
inventory is full message of exchange items

koraykarakus:
## Pull request type

- [x] Bugfix
- [x] Feature
- [ ] Build related changes

Currently, user is not warned when tries to exchange abbys gems and chests while the inventory is full.

## What is the new behaviour?

User now warned with message when he/she tries to exchange item while having no space.

Because this is original behavior and it improves user experience. Packet which we already have is used. Does not involve new logic or feature. Related packet name is WIZ_ITEM_GET.

This PR can solve issue #192 

## Checklist

### DIFF
--- a/Server/GameServer/ItemHandler.cpp
+++ b/Server/GameServer/ItemHandler.cpp
@@ -770,9 +770,16 @@ bool CUser::RunExchange(int nExchangeID, uint32_t count)
 		|| isMerchanting())
 		return false;
 
+	// no empty space in inventory
+	if (!CheckExchange(nExchangeID)) 
+	{
+		Packet result(WIZ_ITEM_GET);
+		result << uint8_t(0); //error code : loot no room = 0
+		Send(&result);
+		return false;
+	}
+
 	if (pExchange == nullptr
-		// Is it a valid exchange (do we have room?)
-			|| !CheckExchange(nExchangeID)
 			// We handle flags from 0-101 only. Anything else is broken.
 			|| pExchange->bRandomFlag > 101)
 			return false;


twostars:
I probably need to adjust the wording in the PR template, but:
> ~~For client changes,~~ I have checked to make sure that this doesn't introduce incompatible behaviour with the official 1.298 server (e.g. unofficial opcodes).

Are you sure this is official behaviour? Does this packet get sent here when using the official server?

koraykarakus:
Must be, i remember seeing that message in game but ofc i don't remember the patch. But around 1299 cypher server in Usko was fresh and everything was just fine. But still cannot confirm that, even searching google does not bring accurate results about it but it improves user experience with UI.

twostars:
Check with the official 1.298 server.

koraykarakus:
what is official 1298 server, there is no such thing exists as far as i know, there are only some private servers.

twostars:
They don't appear to have reuploaded the originals, just the tampered with stuff but:
https://ko4life.net/topic/809-server-files/

The official 1.298 server files.

koraykarakus:
The community and forums appear inactive. I would suggest focusing on developing the game in a way that meets the needs of an average player, so it can serve nostalgia and be used for local or personal purposes. However, if the goal of the project is to deliver an exact replica of the 1298/1299 patch that is fully compatible with everything the original server had, then that’s understandable.

twostars:
I was linking you to a place where you could access the official 1.298 server files to use as a reference for what's official and what's not. Although they are tampered with a little (I didn't see a repost of the original stuff), it's still good for reference.

Issue #180, opened 2025-05-22
Wrong button behavior while displaying character delete is not allowed message


koraykarakus:
Alert box prompts for input when user tries to delete a character, despite a warning that deletion is not allowed.

UTengine:
I prefer the old way tbh.
If anyone did a spastic move while being high or disoriented not paying attention the chance of losing a character is just too high. I assume they disabled this due to people getting hacked aswell.

twostars:
This was resolved by #209

PR #209, opened 2025-05-28
btn behavior for character delete changed

koraykarakus:
## Pull request type

- [x] Bugfix
- [ ] Feature
- [ ] Build related changes

Character delete requesting input from user although it is showing "character delete is disabled right now" message.

## What is the new behaviour?

button behavior changed to BUTTON_CANCEL, now it does not ask for user input, just inform user about the current state of character deletion process.

This was misleading, if current patch of the game has such speciality to remove character from game, button type can be switched between BUTTON_CANCEL and BUTTON_CHARACTER_DELETE by defining and checking constant in GameDef.h in the future updates.

This is also related to issue #180 

## Checklist

- [ ] For client changes, I have checked to make sure that this doesn't introduce incompatible behaviour with the official 1.298 server (e.g. unofficial opcodes).

### DIFF
--- a/Client/WarFare/UICharacterSelect.cpp
+++ b/Client/WarFare/UICharacterSelect.cpp
@@ -114,7 +114,7 @@ bool CUICharacterSelect::ReceiveMessage(CN3UIBase* pSender, uint32_t dwMsg)
 		{
 			std::string szMsg;
 			CGameBase::GetText(IDS_CONFIRM_DELETE_CHR, &szMsg);
-			CGameProcedure::MessageBoxPost(szMsg, "", MB_YESNO, BEHAVIOR_DELETE_CHR);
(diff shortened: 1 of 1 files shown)


twostars:
```cpp
					BEHAVIOR_CANCEL,		// 취소.. 현재는 아이템 거래 신청시 유저가 취소를 눌렀을때..
```
Translates to:
```cpp
// Cancel.. Currently, when the user clicks cancel when requesting an item trade..
```

It should just be BEHAVIOR_NOTHING like everything else, which is actually the default so you could just remove the arg entirely.

koraykarakus:
better, did not translate but i have checked both, they are 100% same, maybe we need a button_behavior in which only OK button exists. Those have OK and cancel which does the same thing (nothing).

koraykarakus:
maybe we can also change MB_YESNO to MB_OK, but don't know original behavior.

Issue #43, opened 2017-04-04
There is no Item and Skill cooldown in Hotkey Window


Demircivi:
![image](https://cloud.githubusercontent.com/assets/22484255/24677701/5966c786-1990-11e7-898d-a5365c7eb8dc.png)

twostars:
This is meant to be handled by Misc\RecastObject\*.dxt

twostars:
This issue was moved to ApexKO/OpenKO#43

PR #210, opened 2025-05-29
Hotkeys now display cooldown timer over them when used.

0xJustMeh:
## Pull request type

- [x] Bugfix
- [ ] Feature
- [ ] Build related changes

No cooldown visual displayed on top of skills/items in hotkey bar

## What is the new behaviour?

Cooldown now renders for skills/items

Fix for issue #43 
-Added rendering logic in UIHotKeyDlg.
- UIHotKeyDlg will check the cooldowns in MagicSkillManager maps and render the cd for the current skillbar. 

## Checklist

### DIFF
--- a/Client/WarFare/MagicSkillMng.cpp
+++ b/Client/WarFare/MagicSkillMng.cpp
@@ -1307,14 +1307,14 @@ void CMagicSkillMng::SetSkillCooldown(__TABLE_UPC_SKILL* pSkill)
 	// Convert iReCastTime (stored in tenths of seconds) to seconds
 	const float fCooldown = static_cast<float>(pSkill->iReCastTime) / 10.0f;
 
-	//if (pSkill->iSelfAnimID1 > 0)
-	//{
+	if (pSkill->iSelfAnimID1 > 0)
+	{
 		m_RecastTimes[pSkill->dwID] = fCooldown;
-	//}
-	//else
-	//{
-	//	m_NonActionRecastTimes[pSkill->dwID] = fCooldown;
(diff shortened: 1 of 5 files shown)


0xJustMeh:
@twostars Looking back, i agree. Since this pr's main goal is to just reference the bug fix, I can slim it down to only that.

> This implementation, though, will only render the one you clicked.

The way it's implemented now will work for any number of the same skill on the bar at the same time, even through different pages. In `CMagicSkillMng::Tick()` I am looping through every skill on every skill bar and updating that skill's `fCurrentCooldDown` with the skill that _should_ be on cooldown. I agree this wasn't ideal either but my thinking was to not couple the UIHotKeyDlg and the MagicSkillManager anymore than they had to be. I did have several implementations before this one where there was more of a lookup list of skills on cd instead though. If we'd rather keep track of cooldowns more like that I'm more than happy to change it.

> While not implemented here, there's actually official logic involving both its targeting behaviour and general usage,

I was trying to think off the top of my head for a reason to have these separated but nothing seemed obvious at the time. I haven't seen much client source code other than this project so I was unaware. I can update this.

twostars:
Ah, I didn't even see the outer context there. Yeah no that's quite horrible; scanning 64 times for every skill in the cooldown map every frame is far from ideal.

The way it officially works has nothing to do with the icons itself, but rather it modifies the placeholder cooldown time (specifically for this purpose) in the skill's struct itself. It's *also* not the most ideal, but still means we're not trying to go out of our way to duplicate this for every available icon.

PR #212, opened 2025-06-01
feature/Options-Dynamic-Resolutions Adds the ability for the Options …

kenner2:
Adds the ability for the Options application to load a list of landscape resolutions supported for the primary monitor:
{7FCF8461-C6FE-4931-A05F-636BC54260B0}

## Pull request type

- [ ] Bugfix
- [x] Feature
- [ ] Build related changes

The Options application currently has a hard-coded list of resolution options, most of which are outdated for modern monitors.

## What is the new behaviour?
The Options application will populate the resolution options based on available resolutions for the primary monitor.  This list will filter out portrait resolutions as well as duplicate entries from the Windows API (there are many options for the same resolution based on the refresh rates available to a given monitor). 

If this fails or returns no entries, make use of the existing hard-coded list as a fail safe.

The list will be sorted with the highest resolutions listed at the top, as many resolution options are available and a user would most likely want one of the higher resolutions.

Why:
I saw the TODO in OptionsDlg.cpp and thought it wouldn't be that hard to implement.

How:
Using the Windows API to identify the primary monitor, get its available resolutions and filter to a usable list.

## Demo
{33475A5C-09DE-4C19-A028-F199B7C9D440}

{6A746489-D29F-4102-962C-578BA92889AC}

## Checklist

### DIFF
--- a/Client/Option/OptionDlg.cpp
+++ b/Client/Option/OptionDlg.cpp
@@ -4,6 +4,10 @@
 #include "stdafx.h"
 #include "Option.h"
 #include "OptionDlg.h"
+#include "Windows.h"
+#include "algorithm"
+#include "set"
+#include "vector"
 
 #ifdef _DEBUG
 #define new DEBUG_NEW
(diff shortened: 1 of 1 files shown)


twostars:
> Github isn't aligning some of the whitespace correctly. Everything is neat within VS Studio '22. The IDE uses space-formatting (tabs generate spaces) whereas the original source used the Tab character (\t). I can look into configuring VS Studio's formatting if it's an issue, just let me know.

It should be using `.editorconfig` from the root folder which *should* be overriding it back to tabs. I don't know why that isn't the case for you.

kenner2:
Not exactly sure what it was, but when I opened VS back up and checked it was using that file.  Re-ran the formatting, seems good now.

PR #213, opened 2025-06-02
1299-1298 login ui

koraykarakus:
## Pull request type

- [x] Bugfix
- [x] Feature
- [ ] Build related changes

1064 login ui, 1064 server select, notice boxes not visible, no packet read for notice texts

## What is the new behaviour?

1298/1299 login ui, server list & notice texts

## Demo

ko1
ko2
ko3

## Checklist


Client/WarFare/GameProcLogIn.h
```diff
@@ -14,23 +14,21 @@
...
-	class CN3Texture*	m_pTexBkg;
 	class CUILogIn*		m_pUILogIn;
-
-	class CN3Camera*	m_pCamera;
-	class CN3Light*		m_pLights[3];
-
+	
 	bool			m_bLogIn; // 로그인 중복 방지..
 	std::string		m_szRegistrationSite;
+	const static uint8_t	m_iMaxServer = 20; //maximum number of servers in .uif file
```
twostars: This seems exclusively used in CUILogin, so I would recommend moving it there instead.
Additionally, it's extremely misleading when it's named m\_\<xyz\> as it's indicating it's a regular member variable, but it's actually static (which would normally be named s\_\<xyz\> with this naming scheme).

Having said that, I would (after moving it to CUILogin) just make it:
```
static constexpr int MAX_SERVERS = 20;
```

Client/WarFare/PacketDef.h
```diff
@@ -10,6 +10,7 @@ const int SOCKET_PORT_LOGIN = 15100;
 #define N3_ACCOUNT_LOGIN			0xf3
 #define N3_ACCOUNT_LOGIN_MGAME		0xf4	// !!! LOGIN_REQ !!! 와 동일
 #define N3_GAMESERVER_GROUP_LIST	0xf5	// Recv - b1(ServerCount) loop(	s1(IP길이), str1(IP문자열), s1(서버이름길이), str1(서버이름문자열), s1(서버접속인원) )
+#define N3_NOTICE_TEXT				0xf6
```
twostars: I would probably go with the more conventional "news" rather than "notice text" because it's confusing as compared to the regular notice packet. Ideally we should just be pulling in the opcodes from the servers and using those, which you'll see are:
```cpp
	PacketHandlers[LS_NEWS]					= &LoginSession::HandleNews;
```

So it'd be keeping it consistent with the naming there.

Client/WarFare/UILogin.cpp
```diff
@@ -34,18 +34,33 @@ CUILogIn::CUILogIn()
 	m_pBtn_Cancel = NULL;
 	m_pBtn_Option = NULL;
 	m_pBtn_Join = NULL;
+	
+	m_pGroup_Notice_1 = NULL;
```
twostars: I see the rest of the old code is using NULL, but this should be updated to use C++11's nullptr instead.
There's no real reason to perpetuate this.

Client/WarFare/UILogin.cpp
```diff
@@ -34,18 +34,33 @@ CUILogIn::CUILogIn()
...
+
 	m_pText_Rights = NULL;
 	m_pImg_MGameLogo = NULL;
 	m_pImg_DaumLogo = NULL;
 	m_pImg_GradeLogo = NULL;
 
-	m_pList_Server = NULL;
+	m_iSelectedServerIndex = -1;
+
+	for (size_t i = 1; i <= CGameProcLogIn::m_iMaxServer; i++) m_pServer_Group[i - 1] = NULL;
```
twostars: These can be done in one loop, and restructured to:
```cpp
for (int i = 0; i < CGameProcLogIn::m_iMaxServer; i++)
{
...
}
```

Additionally, as mentioned above, `CGameProcLogIn::m_iMaxServer` isn't even referenced in CGameProcLogIn, so it should be moved into CUILogin as suggested previously.

Client/WarFare/UILogin.cpp
```diff
@@ -97,11 +112,52 @@ bool CUILogIn::ReceiveMessage(CN3UIBase* pSender, uint32_t dwMsg)
...
 	}
-	else if (UIMSG_LIST_DBLCLK == dwMsg)
+	else if (UIMSG_STRING_LDCLICK == dwMsg) // double click on string
 	{
-		CGameProcedure::s_pProcLogIn->ConnectToGameServer(); // 고른 게임 서버에 접속
-		return true;
+
+		for (int i = 0; i < CGameProcLogIn::m_iMaxServer; ++i)
+		{
+			if (!m_pServer_Group[i]) continue;
```
twostars: Check against nullptr, don't check it as a bool

Client/WarFare/UILogin.cpp
```diff
@@ -97,11 +112,52 @@ bool CUILogIn::ReceiveMessage(CN3UIBase* pSender, uint32_t dwMsg)
...
+
+			CN3UIString* pStr = (CN3UIString*) (m_pServer_Group[i]->GetChildByID("List_Server"));
```
twostars: Fetch this in Load() and check against the fetched entry, don't scan for it on-demand.

Client/WarFare/UILogin.cpp
```diff
@@ -97,11 +112,52 @@ bool CUILogIn::ReceiveMessage(CN3UIBase* pSender, uint32_t dwMsg)
...
+			}
+		}
+
+	}
+	else if (UIMSG_STRING_LCLICK == dwMsg) //change color on left click
+	{
+
+		for (int i = 0; i < CGameProcLogIn::m_iMaxServer; ++i)
+		{
+			if (!m_pServer_Group[i]) continue;
```
twostars: Check against nullptr, don't check it as a bool

Client/WarFare/UILogin.cpp
```diff
@@ -97,11 +112,52 @@ bool CUILogIn::ReceiveMessage(CN3UIBase* pSender, uint32_t dwMsg)
...
+
+			CN3UIString* pStr = (CN3UIString*) (m_pServer_Group[i]->GetChildByID("List_Server"));
+			if (pStr) pStr->SetColor(0xFFFFFFFF); //white
+		}
+
+		for (int i = 0; i < CGameProcLogIn::m_iMaxServer; ++i)
+		{
+			if (!m_pServer_Group[i]) continue;
```
twostars: Check against nullptr, don't check it as a bool

Client/WarFare/UILogin.cpp
```diff
@@ -97,11 +112,52 @@ bool CUILogIn::ReceiveMessage(CN3UIBase* pSender, uint32_t dwMsg)
...
+
+	}
+	else if (UIMSG_STRING_LCLICK == dwMsg) //change color on left click
+	{
+
+		for (int i = 0; i < CGameProcLogIn::m_iMaxServer; ++i)
+		{
+			if (!m_pServer_Group[i]) continue;
+
+			CN3UIString* pStr = (CN3UIString*) (m_pServer_Group[i]->GetChildByID("List_Server"));
```
twostars: Fetch this in Load() and check against the fetched entry, don't scan for it on-demand.

Client/WarFare/UILogin.cpp
```diff
@@ -97,11 +112,52 @@ bool CUILogIn::ReceiveMessage(CN3UIBase* pSender, uint32_t dwMsg)
...
+			if (pStr) pStr->SetColor(0xFFFFFFFF); //white
```
twostars: You can use D3DCOLOR_XRGB() here. It makes it clear the ordering no matter what it's built for or what the colour format is intended to be.

Client/WarFare/UILogin.cpp
```diff
@@ -97,11 +112,52 @@ bool CUILogIn::ReceiveMessage(CN3UIBase* pSender, uint32_t dwMsg)
...
+		}
+
+		for (int i = 0; i < CGameProcLogIn::m_iMaxServer; ++i)
+		{
+			if (!m_pServer_Group[i]) continue;
+
+			CN3UIString* pStr = (CN3UIString*) (m_pServer_Group[i]->GetChildByID("List_Server"));
+			if (pSender == pStr)
+			{
+				pStr->SetColor(0xFFFFFF00); // A=255, R=255, G=255, B=0, yellow
```
twostars: You can use D3DCOLOR_XRGB() here. It makes it clear the ordering no matter what it's built for or what the colour format is intended to be.

Client/WarFare/UILogin.cpp
```diff
@@ -127,46 +183,85 @@ bool CUILogIn::Load(HANDLE hFile)
...
+	m_pStr_Premium->SetVisible(false);
+
+	m_pGroup_ServerList = GetChildByID("Group_ServerList_01");	
+	__ASSERT(m_pGroup_ServerList, "NULL UI Component!!");
+	
+	m_pGroup_ServerList->SetVisible(false);
+
+	
+	//get List_Server (structure : Group_ServerList_01 -> server_20 -> List_Server )
+	for (size_t i = 1; i <= CGameProcLogIn::m_iMaxServer; i++)
```
twostars: I think it's more consistent to go the other way: 0 < max, and then + 1 for the ones that require it.

Client/WarFare/UILogin.cpp
```diff
@@ -127,46 +183,85 @@ bool CUILogIn::Load(HANDLE hFile)
...
-	if (m_pImg_DaumLogo) m_pImg_DaumLogo->SetVisible(false);
+		std::string strPathImage = "img_arrow" + std::to_string(i);
+		CN3UIBase* pChildImage = m_pGroup_ServerList->GetChildByID(strPathImage);
 
-	GetChildByID("Group_Notice_1")->SetVisible(false);
-	GetChildByID("Group_Notice_2")->SetVisible(false);
-	GetChildByID("Group_Notice_3")->SetVisible(false);
-	GetChildByID("premium")->SetVisible(false);
+			
+		if (!pChild || !pChildImage)
```
twostars: Check against nullptr, don't check as a bool

Client/WarFare/UILogin.cpp
```diff
@@ -243,26 +337,69 @@ bool CUILogIn::ServerInfoGet(size_t iIndex, __GameServerInfo& GSI)
...
-	m_pList_Server->ResetContent();
+	
 	if (!m_ListServerInfos.empty())
 	{
 		//sort(m_ListServerInfos.begin(), m_ListServerInfos.end(), not2(__GameServerInfo()));
 		int iSize = m_ListServerInfos.size();
-		for (int i = 0; i < iSize; i++)
+
+		//show ui of existing servers
+		int iMaxPlayerCount = 3000;
```
twostars, line 359: These should not be hardcoded here.

Client/WarFare/GameProcLogIn.cpp
```diff
@@ -32,76 +35,51 @@ static char THIS_FILE[]=__FILE__;
 CGameProcLogIn::CGameProcLogIn()
 {
 	m_pUILogIn	= NULL;
-	m_pChr			= NULL;
```
kenner2: Instead of entirely removing the code to support the orc login screen, could we comment it out and document it as "code to support screaming orc login" or something?  I imagine the original login scene has a lot of issues, but it would still be nice to preserve the supporting code in case someone wanted to mess around with it in the future.

Client/WarFare/UILogin.cpp
```diff
@@ -295,17 +433,178 @@ void CUILogIn::Tick()
 	}
 }
 
+void CUILogIn::AddNoticeText(std::string strNoticeText)
```
twostars: Don't copy strings like this.
If we're doing this, we should use `std::string_view` instead, but if we're using `std::string`, pass it by (const) reference instead, i.e.:
```cpp
void CUILogIn::AddNoticeText(const std::string& strNoticeText)
```

Additionally, as before, in the login server this is called 'news' not 'notice text', so it should be consistent there.

Client/WarFare/UILogin.cpp
```diff
@@ -295,17 +433,178 @@ void CUILogIn::Tick()
...
+{
+	//TODO: needs improvement	
+
+	std::string input = strNoticeText;
```
twostars: There is no need to copy it yet another time here; just use `strNoticeText`.

Client/WarFare/UILogin.cpp
```diff
@@ -295,17 +433,178 @@ void CUILogIn::Tick()
...
+	std::string pieces[10];
```
twostars: Why is this 10? This seems arbitrary, but it supports only 3 boxes, so from your logic, 6 entries.
If it must be hardcoded (but wouldn't it be better to just make it dynamic at that point?), use a constant.

Client/WarFare/GameProcLogIn.cpp
kenner2: I'll be at work for the next couple of hours but I'd be happy to help brainstorm a refactor that preserves the original scene/code but defaults to the newer scene.

I forget when they changed from the orc login screen to the new one.  I want to say it changed when Piana/Breth were replaced with the new Karus/El Morad maps.

Client/WarFare/UILogin.cpp
```diff
@@ -295,17 +433,178 @@ void CUILogIn::Tick()
...
+
+		// jump to next block
+		searchPos = boxEnd + BOX_END_LEN;
+	}
+
+	int iReqNoticeBox = (count + 1) / 2;
+
+	if (iReqNoticeBox <= 1)
+	{
+		CN3UIBase* pNoticeBox = GetChildByID("Group_Notice_1");
```
twostars: None of these should not be scanned on-demand. Find them in Load() first.

Client/WarFare/UILogin.cpp
```diff
@@ -392,33 +692,72 @@ bool CUILogIn::OnKeyPress(int iKey)
...
+			m_iSelectedServerIndex--;
+			
+			uint8_t iServerCount = m_ListServerInfos.size();
+
+			if (m_iSelectedServerIndex < 0)
+				m_iSelectedServerIndex = iServerCount - 1;
 
-			int iIndex = m_pList_Server->GetCurSel();
+			//change its color 
+			CN3UIString* pStrServer = (CN3UIString*) m_pServer_Group[m_iSelectedServerIndex]->GetChildByID("List_Server");
```
twostars: Again: find in Load(), then use that pointer. Don't scan for these on-demand.
These comments go for all of the following cases which are the same.

Client/WarFare/UILogin.cpp
```diff
@@ -392,33 +692,72 @@ bool CUILogIn::OnKeyPress(int iKey)
...
 	{
+
 		switch (iKey)
 		{
 		case DIK_UP:
 		{
-			if (NULL == m_pList_Server) return false;
+			m_iSelectedServerIndex--;
+			
+			uint8_t iServerCount = m_ListServerInfos.size();
```
twostars: There's a few things here:
1. This implicit cast will trigger a warning.
2. The 'i' here indicates it's integral and more importantly, signed. This is unsigned.
3. The logic below immediately -1s it.
If this is signed (which it is not -- you've made it unsigned), and the list is empty, this will revert it to -1 which -- besides the fact there's no sanity checks there, would ordinarily consider it a valid value.
Since this is unsigned, when the list is empty this becomes 255, which causes an overflow.
Just keep it int, like the rest of it.

Client/WarFare/UILogin.cpp
twostars: Additionally, these should be sanity checked. It shouldn't assume that it's valid.
As-is, if there's no servers, it will overflow (or if it was signed above, it would do the opposite). It should be sanity checked.

Client/WarFare/UILogin.cpp
```diff
@@ -392,33 +692,72 @@ bool CUILogIn::OnKeyPress(int iKey)
...
+				m_iSelectedServerIndex = iServerCount - 1;
 
-			int iIndex = m_pList_Server->GetCurSel();
+			//change its color 
+			CN3UIString* pStrServer = (CN3UIString*) m_pServer_Group[m_iSelectedServerIndex]->GetChildByID("List_Server");
+			__ASSERT(pStrServer, "NULL UI Component!!");
 
-			if (iIndex > 0) m_pList_Server->SetCurSel(iIndex - 1);
-			int iCnt = m_pList_Server->GetCount();
+			this->ReceiveMessage(pStrServer, UIMSG_STRING_LCLICK);
```
twostars: We can drop 'this->'. It's totally unnecessary.

Client/WarFare/UILogin.h
```diff
@@ -56,29 +56,46 @@ class CUILogIn : public CN3UIBase
...
+
+
 	CN3UIBase* m_pText_Rights;
 	CN3UIBase* m_pImg_MGameLogo;
 	CN3UIBase* m_pImg_DaumLogo;
 	CN3UIBase* m_pImg_GradeLogo;
 
-	CN3UIList*	m_pList_Server;
-	
+	CN3UIBase* m_pServer_Group[20];
```
twostars: These should be using a constant (preferably the existing constant you have, just, again, moved here into CUILogin).

twostars (dismissed):
Just did a cursory look over the code. Could use a bit of cleanup, but it's not too bad.
Just a few things that really need to be changed before I can merge it in.

Client/WarFare/GameProcLogIn.cpp
twostars: It's a bit of a slippery slope to be honest, but at that point it's best to just maintain 2 different versions of CGameProcLogin and CUILogin, otherwise it complicates things far too much.

The newer stuff is actually officially named differently to begin with; CGameProcIntro and CUILoginIntro.

Client/WarFare/GameProcLogIn.cpp
```diff
@@ -32,76 +35,51 @@ static char THIS_FILE[]=__FILE__;
...
+	m_pUILogIn->LoadFromFile(szPath);
+
+	s_pEng->s_SndMgr.ReleaseStreamObj(&(CGameProcedure::s_pSnd_BGM));
 
-	// TODO(srmeier): the UIE saves in the old format...
-	if (pTbl) m_pUILogIn->LoadFromFile(pTbl->szLogIn);//, N3FORMAT_VER_1298);
+	//NOTE: mp3 files don't work
+	//this is used for testing. original sound is not in Snd folder.
+	//it exists as bgm_el_battle.mp3, convert it to .wav to make it work
+	std::string szFN = "Snd\\bgm_el_battle.wav";
```
kenner2: Do we have a constants file for string values?  Looks like resources.h is entirely numeric mappings.

Nothing that needs to be solved as a part of your PR, just thinking out loud:
Constant files for strings can be great for code traceability.  It can make it very easy to track down all uses of a resource/phrase and gives a location to browse string data.

Thinking forward, it would also help set up for something like resource bundle overrides.  This is something I was kicking around in my head last night and I'll do a separate write-up (as an issue, I guess?) to gather thoughts, but the TLDR would be:
1.  Have a file that contains your default constants/string data that is checked into git.  Changing a default changes it for the baseline.
2.  You have an override file on .gitignore that a dev can use to change those values locally without affecting the mainline.  This allows some level of customization for _their_ build without affecting anyone else.
3.  You have a template of the override file that is checked into git for developer reference, but isn't actively used in the build process.

This kind of relates to what I was thinking in terms of preserving the screaming orc scene.  For something like that, a solution would be closer to the dependency injection design pattern rather than resource bundling.

twostars: I missed this line. For music, these are unfortunately officially hardcoded filenames. I think we should just update audio to support the official assets (and hence MP3) though; we shouldn't expect custom changes here as we're otherwise using the official assets. Better to use request MP3 and have it fail to load (and not have to change it later) than have it WAV, expect the user to change it, then have to swap it later, and expect the user to change the asset back later.
Better it just doesn't load period until MP3 is supported (or we use more modern assets because libogg is opensource).

As far as overriding constant strings, most strings are loaded from the official TBL data (which are official client assets), which is where the resources.h mappings relates to.

While those should probably be separated from resource.h (since they're not loaded from PE resources anymore, as they were originally for a while sometime before 1.298), I don't really understand the use case here. You can just change the resource in the TBL in this case, so long as the format is the same, no?

koraykarakus: it is normally taken from sound.tbl
but inside sound.tbl it exists as .mp3 file bgm_el_battle.mp3
and soundmanager does not work with mp3 files, i just converted ( changed extension of mp3 to wav not real convert ) and tested it. It is again battle music and should be changed. Client data folders missing real login music which is

https://www.youtube.com/watch?v=ZVgq8609LKE

that music should be added as real wav into client.

twostars: Not for login; at least, not how they use them. They're hardcoded officially.

image

kenner2: I'm chaining too many thoughts into a post.  Should really just boil this down to:

Using a constants file for strings is good code management in most cases.  If, for some reason, a constant needs to be updated/changed/fixed, you can easily track down everywhere it's used (potentially turning up and fixing related bugs) and update it one location. 

I'll do a more coherent write-up on DI thoughts at some other point.  You're right that it wouldn't make sense for actual constants like file paths/names, and that local TBL files are existing resource bundles for the client.

twostars: Since there are some (but only a few) in the client assets, we could add a hack for swapping the extension for MP3 in the sound manager temporarily until it's supported. That way we wouldn't need to change the outer stuff; just remove the hack later.

But the fact it's not even the right background music is another issue that I hadn't noticed until now either. It should really just be swapped to the real filename, regardless of its ability to play it.

Client/WarFare/GameProcLogIn.h
koraykarakus: this is done, you can check commit 2

Client/WarFare/PacketDef.h
koraykarakus: done in commit 3.

Client/WarFare/UILogin.cpp
koraykarakus: done in commit 4.

Client/WarFare/UILogin.cpp
koraykarakus: done in commit 5.

Client/WarFare/UILogin.cpp
koraykarakus: this is done.

Client/WarFare/UILogin.h
koraykarakus: this is also done

Client/WarFare/GameProcLogIn.cpp
```diff
@@ -32,76 +35,51 @@ static char THIS_FILE[]=__FILE__;
...
-	m_pLights[1]->LoadFromFile("Intro\\1.N3Light");
-	m_pLights[2]->LoadFromFile("Intro\\2.N3Light");
-
-	s_pEng->s_SndMgr.ReleaseStreamObj(&(CGameProcedure::s_pSnd_BGM));
-	CGameProcedure::s_pSnd_BGM = s_pEng->s_SndMgr.CreateStreamObj(35);	//몬스터 울부짖는 26초짜리 소리..
+	//random elmorad or karus background
+	std::string strNationPrefix = "el";
+	std::srand(std::time(nullptr));
+	int iRand = std::rand() % 101;
+	(iRand < 50) ? strNationPrefix = "el" : strNationPrefix = "ka";
```
kenner2: With the NULL vs nullptr thing being pointed out as a code style convention, what do you think about ternary operators @twostars ?  I don't mind them but I know a lot of devs dislike them.  The general complaint is understandable:  An if/else statement is more human readable and is easier to understand, even if it takes up a bit more space.

kenner2: Also, if we're keeping the ternary, then a nitpick on the line:  The goal of the line is to assign a value to strNationPrefix.  You can make that obvious by assigning it the return value of the ternary operation:

strNationPrefix = iRand < 50 ? "el" : "ka";

Client/WarFare/GameProcLogIn.cpp
koraykarakus: see this to keep both login screens
#215

Client/WarFare/GameProcLogIn.cpp
twostars: Yeah, it should ideally just be a standard if statement here, or as you say, move the ternary to the assignment.
That having said, this is pointless anyway because it should be pulling the name from the TBL.

UTengine:
Just as a side note you might want to put srand time and based on that load karus or Eldorad login otherwise it will always just display the same nation. That's how I wrote the 1298 login
Example:
```
#include <bits/stdc++.h>
using namespace std;

int main() {

    // Use current time as seed for random generator
    srand(time(0));
    for (int i = 0; i < 3; i++)
        cout << rand() << endl;
    return 0;
}
```

```
#include <ctime>  // Include ctime to access the current time

int main() {
    std::srand(std::time(0));  // Seed the random generator with current time
    int random_value = std::rand();  // Generate a random number
    std::cout << "Random Number: " << random_value << '\n';
    return 0;
}
```

UTengine:
It would be a nice addition if you added | as a linebreak to CN3UIString, and then in the logins enver.ini have 3 news lines that's just one text but each line would be separated by | so the logins enver would send
Notice1=Line1Hi|Line2Yo|Line3Bye
Notice2=....
Notice3=....
@koraykarakus

twostars:
> It would be a nice addition if you added | as a linebreak to CN3UIString, and then in the logins enver.ini have 3 news lines that's just one text but each line would be separated by | so the logins enver would send Notice1=Line1Hi|Line2Yo|Line3Bye Notice2=.... Notice3=.... @koraykarakus It's just a nlength packet with std strings..

For the record, this isn't official behaviour with news at all, let alone CN3UIString. It's officially implemented in very niche cases, e.g. for quest_talk dialogue.
This is very off-track now but you could do it on the server's side and translate that back to \n, which would then be handled by the client appropriately (assuming the control is even styled for multiline in the first place).

koraykarakus:
@UTengine  can't answer back to your message so i send new one. On server side string written very confusing. It starts writing packet by adding "Login Notice", after that it adds full length, puts "#", and continue with message + length (Or something similar to that ). While reading it lengths should be skipped because they have written inside string.

Editing server side packet sender and simplfying it would be better option, it can just send each title with their lengths and each message with their lengts in a well organized way. 

for now you can use them as

[NEWS]
TITLE_00=title0
MESSAGE_00=message0
TITLE_01=title1
MESSAGE_01=message01
TITLE_02=title2
MESSAGE_02=message02

inside /bin/ LoginServer.ini

PR #215, opened 2025-06-02
Login UI double version

koraykarakus:
## Pull request type

- [ ] Bugfix
- [x] Feature
- [ ] Build related changes

current PR support only 1299 login screen 

## What is the new behaviour?

supports both versions

this caused so many changes so I decided to open new PR for it

## Checklist

### DIFF
--- a/Client/WarFare/GameDef.h
+++ b/Client/WarFare/GameDef.h
@@ -18,8 +18,8 @@
 #endif
 #include "shared/Packet.h"
 
-constexpr int CURRENT_VERSION = 1298;//1068; // 현재 버전
-
+constexpr int CURRENT_VERSION = 1298;	//1068; // 현재 버전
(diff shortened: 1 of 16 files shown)


kenner2:
For the short term I think it might be better just to take the original versions of the GameProcLogin and UILogin files and add them to your original PR with a .old extension (GameProcLogin.h.old, GameprocLogin.cpp.old, etc).  That way we have the original code archived and we can revisit being able to build either screen.  

The way you have this structured probably works well, but it'll build both sets of logic into the exe, regardless of which the developer intends to use for their server.  I think with a bit of thought and build management we can have a build configuration file that drives which login screen object gets used in the build, so only the desired set of functionality is built into the exe

Just my two cents; I'm happy as long as we're not losing that original code.

twostars:
Just keep it to the other PR. Don't worry about preserving it for now; just focus on the other stuff.

PR #216, opened 2025-06-03
kasar hood ID written wrong, causes it not to be closed

koraykarakus:
## Pull request type

- [ ] Bugfix
- [ ] Feature
- [ ] Build related changes

current class names on skill tree for el morad is not visible properly because kasar hood ( master rogue ) text is not closed correctly. Hence, for a warrior without master, class name seen as kasar hood, although it should be Blade.

## What is the new behaviour?

Now, it is seen as Blade as it should be.

ID written wrong in func SetPageInCharRegion(), in SkillTreeDlg.cpp. 

## Checklist


twostars:
Rather than this, to be consistent with everything else, we should just make the string comparison case-insensitive.

PR #219, opened 2025-06-04
Anvil UI added

Mervandeli:
## Pull request type

- [ ✓] Bugfix
- [✓ ] Feature
- [ ] Build related changes

-UpgradeSelect UI show "CUITemUpgrade and Ring Upgrade Should be implemented"

## What is the new behaviour?
-ItemUpgrade Window being  displayed And the inventory items were retrieved correctly
-ItemUpgrade Window being  displayed And the inventory items were retrieved correctly

 I made these changes to integrate the UIItemUpgrade and UIRingUpgrade classes without breaking the existing structure. In order to make them functional, I had to modify the corresponding .uif files—co_itemupgrade.uif and co_ringupgrade.uif. I'm not certain whether these files were based on an older version, but the modifications were necessary to properly link the new classes with the UI. After the changes, the Anvil UI displayed correctly, and the inventory items were retrieved properly. However, further work is needed in the ReceiveIconDrop method, where the drag-and-drop logic should continue. I'm planning to work on that part collaboratively.

## Demo
In co_itemupgrade.uif and co_ringupgrade_uif , I converted the ID_string values from a_slot_0 to a_slot_28 into numeric values ranging from 0 to 28 accordingly.

Ekran görüntüsü 2025-05-31 131039
Ekran görüntüsü 2025-06-03 120402

## Checklist


UTengine:
Good job I already made the anvil but had to clean it up before doing a pull request. I don't mind, the only issue is if I were to add my code to this I would have to read everything from 0 to add the missing features and I don't feel like doing so. I expected it to be finished when I took a look. I don't see a tick I don't see receivemessage with the packet results or I skipped too fast when going through the commit.

twostars:
In that case, I'm unfortunately forced to close the PR.

The aim is to keep things compatible with the 1.298 server and client, until such a time as there's no need for further testing against these. Changing the UIs breaks them for the official 1.298 client, meaning they cannot be tested side-by-side on the same data.

PR #221, opened 2025-06-05
Master quests for all classes & nations

koraykarakus:
## Pull request type

- [x] Bugfix
- [x] Feature
- [ ] Build related changes

At current state, some required functions for master quests to work is not added on server side ( lua and cpp files ). Those functions are : CheckPromotionEligible, GivePromotionQuest. Therefore, master quests are not working at all. User cannot get any quest from NPC also user cannot return and promote his/her character to master.

## What is the new behaviour?

Mentioned functions declared and filled. They are now working fine. User can get master quests ( all classes, warrior, rogue, priest, mage & all nations ), can return quests to be promoted to master. Required items from user has been robbed from his/her inventory.

This is one of main features of the game and it should be working fine. Required additions done to unfinished parts of QuestHandler and lua scripts.

## Demo

Tested with all classes and nations. Both under 60 and above levels and with items in inventory or without items in inventory.
## Checklist

### DIFF
--- a/Server/GameServer/QuestHandler.cpp
+++ b/Server/GameServer/QuestHandler.cpp
@@ -152,12 +152,100 @@ bool CUser::PromoteUserNovice()
 // From novice to master.
 bool CUser::PromoteUser()
 {
-	/* unlike the official, the checks & item removal should be handled in the script, not here */
-	uint8_t bOldClass = GetClassType();
-
-	// We must be a novice before we can be promoted to master.
-	if (!isNovice()) 
+	//re-check if user really eligible
+	if (CheckPromotionEligible() != 1)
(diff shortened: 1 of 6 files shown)


twostars:
See this comment here that you removed:
https://github.com/Open-KO/KnightOnline/pull/221/files#diff-a4529ce24301b5e97f58f40fe91e454ab30e6d90cb760995221a08de49391c9fL155

The quest stuff is all inadvertently replaced with Lua anyway because we skipped straight to using snoxd-koserver's newer stuff which ended up downgraded for 1.298, so there's not really any point here to forcing arbitrary hardcoded item requirements in the server.
Just do it in the script.

PR #222, opened 2025-06-05
max level limit in seeking party text changed to 80

koraykarakus:
## Pull request type

- [x] Bugfix
- [ ] Feature
- [ ] Build related changes

Max level of seeking party text has been set as 60.

## What is the new behaviour?

It has been set as 80 which is the actual max level for 1299 version.

This was causing misinformation for high level characters for example if a 79 level user opens seeking party feature, he/she sees level range as 71-60 and that is wrong.  

## Checklist


twostars:
Level aside, this is not the only case that needs updating. Search for the other IDS_WANT_PARTY_MEMBER cases.

PR #224, opened 2025-06-06
to buy loyalty from npc knight clerk

koraykarakus:
## Pull request type

- [x] Bugfix
- [ ] Feature
- [ ] Build related changes

feature to buy national points from NPC is not working at the moment.

## What is the new behaviour?

It is working now for both nations. User can buy 100 or 500 np by paying 350000 or 1500000 coins from NPC Knight Clerk located in main cities.

This is an important feature of the game and it should be working.

Note: this addition gives user national points but it is also affected by premium bonuses and it also gives right side national points. Current function under User.cpp works in that way. I don't remember what is official feature, does it really give right side national points ? does it really give bonusses from premium. If so, additional lua method can be added to directly add left side NP.

## Checklist

### DIFF
--- a/Server/bin/QUESTS/1.lua
+++ b/Server/bin/QUESTS/1.lua
@@ -2018,49 +2018,36 @@ elseif nEventID == 15951 then
 	end
 	end
 elseif nEventID == 16951 then
-	pUser:SendDebugString("Unknown LOGIC command 'CHECK_LOYALTY'.");
-	if false then -- unknown logic command (CHECK_LOYALTY)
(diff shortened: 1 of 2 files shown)


twostars:
Mixed feelings on this. These scripts are generated from the original EVTs with the EvtConverter tool here:
https://github.com/Open-KO/KnightOnline/tree/master/Server/EvtConverter

Half the stuff replaced is either auto-generated or unhandled. For consistency's sake (since the EvtConverter is still very much in its infancy) I think I'd prefer the EvtConverter updated, rather than the script. That way we can regenerate these scripts upon update without having to manually preserve changes.

PR #225, opened 2025-06-07
fix user cannot close npctalk menu with close button

koraykarakus:
## Pull request type

- [x] Bugfix
- [ ] Feature
- [ ] Build related changes

Current state : user cannot close NpcTalk menu with close button because close button visibility set as false inside UIQuestTalk.cpp, also it is not initialized and does not define as member.

## What is the new behaviour?

m_pBtnClose initialized in constructor, defined in header file and its visibility set as default ( visible ). By this way, button is visible now and user can close NpcTalk menu by clicking the button.

This change improves user experience, close button should be working, otherwise user is forced to display all of the messages of selected npc.

## Checklist

### DIFF
--- a/Client/WarFare/UIQuestTalk.cpp
+++ b/Client/WarFare/UIQuestTalk.cpp
@@ -26,7 +26,7 @@ CUIQuestTalk::CUIQuestTalk()
 {
 	m_pTextTalk	= NULL;
 	m_pBtnOk	= NULL;
-
+	m_pBtnClose = NULL;
 	m_iNumTalk	= 0;
 	m_iCurTalk	= 0;
 }
@@ -78,6 +78,10 @@ bool CUIQuestTalk::ReceiveMessage(CN3UIBase *pSender, uint32_t dwMsg)
 				m_pTextTalk->SetString(m_szTalk[m_iCurTalk]);
 			}
 		}
+		else if (pSender == m_pBtnClose)
+		{
+			SetVisible(false);
+		}
 	}
 
 	return true;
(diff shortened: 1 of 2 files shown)


Client/WarFare/UIQuestTalk.cpp
```diff
@@ -92,10 +96,9 @@ bool CUIQuestTalk::Load(HANDLE hFile)
 	m_pBtnOk	= (CN3UIButton*)(this->GetChildByID("btn_Ok_center"));		__ASSERT(m_pBtnOk, "NULL UI Component!!!");
 
 	// NOTE(srmeier): new stuff
-	CN3UIButton* m_pBtnClose = (CN3UIButton*)(this->GetChildByID("btn_close"));
-	if (m_pBtnClose) {
-		m_pBtnClose->SetVisible(false);
-	}
+	m_pBtnClose = (CN3UIButton*)(this->GetChildByID("btn_close"));
+	__ASSERT(m_pBtnClose, "NULL UI Component!!!");
```
kenner2: Should have the object name in the error message in case this is ever hit and needs to be researched  (Something like "m_pBtnClose is null").  __ASSERT gives file/line info, but clear error messaging is still good

Client/WarFare/UIQuestTalk.cpp
```diff
@@ -26,7 +26,7 @@ CUIQuestTalk::CUIQuestTalk()
 {
 	m_pTextTalk	= NULL;
 	m_pBtnOk	= NULL;
-
+	m_pBtnClose = NULL;
```
kenner2: Been mentioned in other PRs that we should just use nullptr instead of the NULL macro going forward

twostars (approved):

PR #228, opened 2025-06-08
lua method for quest state updating

koraykarakus:
## Pull request type

- [ ] Bugfix
- [x] Feature
- [ ] Build related changes

Quest state update packet is not send back to user, that is not correct behavior. After accepting each quest, user should be warned with a text message which contains quest name and information (quest state accepted ).

## What is the new behaviour?

SendQuestStateUpdate also defined as lua method, so now it can also be used inside lua scripts. At current state that function is only used for master quests.

User.h and lua_bindings.cpp changed respectively.

## Checklist

### DIFF
--- a/Server/GameServer/User.h
+++ b/Server/GameServer/User.h
@@ -1300,4 +1300,9 @@ public:
 	DECLARE_LUA_FUNCTION(CheckMonsterChallengeUserCount) {
 		LUA_RETURN(LUA_GET_INSTANCE()->GetMonsterChallengeUserCount());
 	}
+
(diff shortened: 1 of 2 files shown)


twostars:
This isn't needed for what you propose. Use SaveEvent() for updating quest states.

It not sending the update here is normal -- for 1298, which is because the update doesn't actually work like it does in newer clients. The "update" only affects the mastery abilities so there's no reason to send it otherwise (and why even officially it's filtered as such). Only the list packet updates the list itself.

Officially, anyway. I'm fairly certain none of this is even updated properly in the client.

Issue #231, opened 2025-06-14
1298 Login Screen Misalignment


kenner2:
The 1298 Login Screen is not anchoring to the center points.  Tried multiple resolutions, full screen and windowed:

1600x1200 Windowed:

![Image](https://github.com/user-attachments/assets/990596bb-5546-4340-a46b-ee66a15160ed)

2560x1440 Full Screen:

![Image](https://github.com/user-attachments/assets/39b5441e-640b-4771-b52e-e593f5969954)

PR #232, opened 2025-06-14
bugfix/#231 1298 Login Screen UI Misaligned at higher resolutions

kenner2:
See:  #231 

## Pull request type
- [x] Bugfix

At higher resolutions (where the height of the application is larger than the default height of the login UI), the 1298 Login UI is not correctly centered on the y-axis.

Further, child elements were not be correctly positioned when calling their position functions (like SetPosCenter()). This is because these elements are positioned in the CUILogIn_1298::Load() function, which is called before the resizing code.  By setting the the CUILogIn_1298 size from the camera inside of the Load() function, all child elements should be correctly positioned.

## What is the new behaviour?
Calculation for y-axis delta is now divided by 2.

To correct the bug reported in #231

## Demo
image

## Checklist


kenner2:
Changed to draft.  Login server list still not vertically aligned:
image

PR #235, opened 2025-06-15
n3tool save function rework

koraykarakus:
## Pull request type

- [x] Bugfix
- [x] Feature

saved file with uie does not open by game because it does not support new versions. New version adds "00 00 00 00" padding after each string can be seen by checking current uif files of game with HxD editor. 

## What is the new behaviour?

Padding after each string also added so saved file can be opened by game.

crucial to change UI bugs. ( for example scrollbar ID is forgotten to be set in ui file of CMDList ) 

## Demo

ui_save_1

## Checklist

### DIFF
--- a/Client/N3Base/N3UIBase.cpp
+++ b/Client/N3Base/N3UIBase.cpp
@@ -676,7 +676,25 @@ bool CN3UIBase::Save(HANDLE hFile)
 
 	// child 정보
 	int iCC = m_Children.size();
-	WriteFile(hFile, &iCC, sizeof(iCC), &dwRWC, NULL); // Child 갯수 ㅆ고..고..
+
+	if (m_iFileFormatVersion >= N3FORMAT_VER_1264)
+	{
+		TRACE("version bigger or equal \n");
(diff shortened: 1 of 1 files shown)


Client/N3Base/N3UIBase.cpp
```diff
@@ -685,6 +703,13 @@ bool CN3UIBase::Save(HANDLE hFile)
 
 		WriteFile(hFile, &eUIType, sizeof(eUIType), &dwRWC, NULL); // UI Type 쓰고..
 		pChild->Save(hFile);
+
+		if (eUIType == UI_TYPE_STRING)
```
twostars: Maybe just write this from CN3UIString::Save()?

twostars (changes requested):

Client/N3Base/N3UIBase.cpp
koraykarakus: carried it, tested also by saving empty string, looks like empty strings also requiring 4bytes of padding.

PR #241, opened 2025-06-23
weight bug fix

koraykarakus:
## Pull request type

- [x] Bugfix

weight info sent as uint16_t from server but read as int16_t at client. Therefore, overflow happens when the user have higher numbers than 3276,7 ( 32767 int16_t max) . 

## What is the new behaviour?

static cast to int after reading as uint16_t so it does not overflow anymore.

For a high level character with maxed STR, game is unplayable. Because of this bug, user cannot buy / trade items.

## Checklist


twostars:
> weight info sent as uint16_t from server but read as int16_t at client. Therefore, overflow happens when the user have higher numbers than 3276,7 ( 32767 int16_t max) .

But *why* do they have such a large weight here? It's officially handled as signed so this shouldn't be necessary; to my understanding, the weight should never exceed this limit. Is this perhaps a server/data bug with the weight?

koraykarakus:
on client side it is divided to 10 to represent weight in decimals. maybe in order not to keep it as float on DB. I remember this bug from game long ago definatelly they read signed on client but they have fixed this bug at some point. This thing was so old breath/piana existing.

PR #242, opened 2025-06-23
glow effect corrections for both unique and normal items

koraykarakus:
## Pull request type

- [x] Bugfix
- [x] Feature

Items glows arbitrary ( set wron paramater as 64 so +7 items also glow ) . Unique items should glow different but they are currently same with normal items. Some normal items has special glowing effect ( unique glow ) like : raptor, totamic spear etc.

## What is the new behaviour?

Unique items glow as expected, some normal items added inside GameDef to glow like unique as mentioned above. Need to add other items which behave same.

An important feature of the game which should be addressed.

## Demo

glow_2
glow_3

## Checklist

### DIFF
--- a/Client/WarFare/GameDef.h
+++ b/Client/WarFare/GameDef.h
@@ -807,6 +807,7 @@ typedef struct __TABLE_ITEM_BASIC // 장착 아이템에 관한 리소스 레코
 
 const int MAX_ITEM_EXTENSION = 24; // Number of item extension tables. (Item_Ext_0..23.tbl is a total of 24)
 const int LIMIT_FX_DAMAGE = 64;
+const int LIMIT_FX_GLOW = 80; //When an item starts to glow
 const int ITEM_LIMITED_EXHAUST = 17;
 
 typedef struct __TABLE_ITEM_EXT // 장착 아이템에 관한 리소스 레코드...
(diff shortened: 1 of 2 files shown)


twostars:
This does not need to be handled via explicit item ID lookups. Rather, look to the item TBL data; all of the information you need is there.

koraykarakus:
I checked and see now these details really exist inside the Item_Org_us.tbl file. However, the values are 1000 times larger than needed, and the FXTailID does not exist at all. Therefore, for each item, two additional calculations must be done in the code: dividing the glow value by 1000 for glow and adding 1 to first calculation for the tail value.

Additionally, some items are missing from the Item_Org_us.tbl file (such as Hell Breaker and Gigantic Axe, if I remember correctly), and some items share the same base ID with others—for example, Gigantic Axe and Tabar, Smite Hammer and another item, Javana Axe and another item. I checked that long ago to provide official appearance for each item but at the end I achieved that after making changes on both database and tbl files. Maybe they can be searched inside code and than be overwritten but still it is not the correct way in my opinion.

The cleanest solution would be to edit the .tbl file to correct all these issues and then adjust the source code accordingly. Some of the mentioned items also share the same base ID in the database. Maybe we shall keep tbl files as text on github and than convert them back to tbl while building project on VS.

twostars:
It's not a data issue... the TBL data just needs to be handled correctly. You don't need to match specific item IDs for this.

koraykarakus:
indeed addresing special ids also wrong but tbl as well. 
datafromtbl / 1000 gives FX address
and datafrom tbl / 1000 + 1 gives other FX address.
some items don't have that data because they are not in item_org_us.tbl ( example: gigantic axe, hell breaker )
doing it like that brings in 2 extra calculations for each item, assuming support for rebirth items and armors this will bring more calculation. Assuming 500+ people standing in zone, that will result in some fps drop ( maybe small but it will ).

on the other hand if effectID2 corrected in tbl and missing items added inside that tbl, no calculations required.

UTengine:
Jesus christ don't do this.
It should be somewhere in the player base or other player files or named similar just search for the preexisting effect id's.
There is a item pointer check somewhere along that file I can't remember just reuse it and add if ->ext->dwItemGlowEffect>0;
Use the glow. Else use existing elemental code calculations but I think those have been removed from game idk either reverse client or ask twostars.
Then for the tail or w.e can do simply itemgloweffect+1
There's really no need to define every fricking glow imho clutters everything up. Also npc's can have item's with gloweffects. All new item effects have to be read from EXT not basic.

koraykarakus:
i ll turn back to editting it someday, either way because of the fact that TBL info is wrong, at some point you need to change TBL. I have encountered some examples long ago but forgot which items they were. In my opinion, just go with tbl , what is the point of adding new calculations to base code, but improvement and for testing purposes it can be overwritten inside code. 

I gave it a quick try by edding enums on GameDef to test actually, after seeing it works PR ed, than I realized they are written inside tbl files, I ll set those and read from tbl when I can in the future.

twostars:
Do you plan on updating this PR?

twostars:
You should probably discuss what you intend to do with it first before getting stuck into it. It shouldn't really affect that much.

koraykarakus:
When i first tried to make it work, i was unaware of official exe exists and it works just fine. Therefore, while working on this PR i included some item IDs and continued work by guess than test. So that, now I have official exe, can use TBLs and make them work like in official. There won't be any revolutionary changes which affect other parts of code but committing on this PR as change is sadly not an option ( requires more work on branch updates, cleaning etc ).

twostars:
You can just reset it and redo your PR with the same branch.

koraykarakus:
dealing with git harder than dealing with C++, it causes a lot other problems most of the time so i avoid Git part

Issue #238, opened 2025-06-16
Should the server project be reset back to base


twostars:
The server project as-is, is a history-less fork of the snoxd-koserver project which was originally based on the leaked source (as is the client source).

It was intended for higher versions than 1.298 (1.886 originally, if I recall correctly), and has seen a lot of changes. Everything between the snoxd-koserver repo & changes in this repo is entirely undocumented, and due to it being shoved in like that from elsewhere, the original history of snoxd-koserver isn't preserved either.

To some extent this has been worked around; @srmeier keeping the Lua implementation and writing the EvtConverter project to convert the EVTs back to Lua scripts (which is incomplete as of this time).

As such, it feels like it's in a very awkward state.

I have to wonder -- is it better at this point to just swap it back to the leaked base and build it up for 1.298, or continue with it as-is?
It feels like doing a lot of the work again, but then again -- it can be kept a little more closely to mgame's original implementation than it ended up.

It just feels like an awkward mess as it is, and my biggest gripe with this repository.

twostars:
I've reverted *most* of the obvious newer stuff that has no place, except Lua support for scripts which I retrofitted better to handle the old behaviour, so 1.886 stuff isn't as big a concern.

The issue is more stuff like, updated/fixed feature support such as skills, or even basic things like trading (outdated and buggy) or merchanting (wouldn't exist at all).

It would be a considerable setback, but not entirely unmanageable. I think it'd give a better indication of what actually needs doing, and keep things better in the scope of official behaviour. It'd also mean we don't have to worry about the random behaviour caused by some fairly nonsense changes introduced randomly between snoxd-koserver and this repo.

I'd probably approach it with a separate branch to get it back to basically running and communicating with the 1.298 client and *then* swap it. It'd be missing features, sure, but it'd still be usable.

kenner2:
>When you say you want it pushed to a branch to diff it, you mean over the top to compare with a git diff, right? A lot of the code was moved around and reworked, and isn't going to directly diff well if that's what you're hoping for.

I mean in a separate folder - or just in a different repo for the sake of archival if you'd like.  I'm not expecting to work magic with an actual diff program

twostars:
To be clear, I started the snoxd-koserver project over a decade ago. It's not currently being maintained, and hasn't for many years. Its main repo doesn't even exist anymore. It has nothing to do with the people who took the domain and call themselves 'snoxd' today.

[...]

PR #248, opened 2025-06-25
Revert server code to base 1.081 & update to 1.298

### LINKED ISSUE #234: Login Server caches TB_User
The Login Server caches the account list from TB_USER on start-up instead of calling the database when a login attempt is made.  This means that if a new account is registered or a password is updated the records will be stale until the Login Server is restarted.

twostars:
As per #238, the goal was to revert the server code back to base 1.081 code and then update it - somewhat minimally - to support 1.298 again.

The previous codebase used a historyless fork of snoxd-koserver (which deviated after snoxd-koserver stopped being maintained, and before OpenKO was a thing), which in itself was originally built from the 1.081 code... but with very different goals in mind.

It aimed to support the newest versions at the time (which was originally 1.886 or so), and didn't have any real desire to remain close to official logic -- it just intended to emulate the official behaviour in a close enough approximation.

This means the codebase was designed for far newer versions, where in some cases features were entirely reworked and not applicable to 1.298 anymore, so we were forced to downgrade as best we could for 1.298. This left us with things like Lua scripts for the quests, and @srmeier's EvtConverter aiming to convert the EVT scripts to Lua (which to date hasn't been finished).

The codebase was also fairly overengineered, and in many ways a lot more difficult for newcomers to understand.

So, the goal was to revert back to 1.081 and redo it -- with the benefits C++20 affords us (and really, back when snoxd-koserver was started, C++11 was barely a thing).
From this point, we can build it up to what we need, which is a server specifically designed for 1.298. We can document as we go, and try to keep things in a nice state before the codebase gets massively inflated.

As this goes back to official logic, this closes #234.

As far as feature support, we only restored basic support for 1.298 server behaviour, so: things like getting in-game, moving around, seeing NPCs, seeing players, chatting, using the warp gate, teleporting to other zones, etc.

There is a plethora of features that are no longer supported (though most weren't supported by the client anyway), but things like skill logic being outdated, EVTs not being fully supported, even /town requiring updating (because it uses START_POSITION, which didn't exist in the base source), will be left for future PRs.

As far as features that require database tables, before we start building on these, we need to figure out what we're doing with them in future.
e.g. Using some API like SQLAPI++ (to throw out a random one), or perhaps having stuff generated via the separate database tooling that's currently in the works (and then the question is, to what extent).

They can be readily built on once that's established. Until then, it's probably better to avoid adding more unnecessary work; there's not that many tables loaded at present, so migrating will be easier.

But as far as it being basically 1.298 ready, it should be as of this PR.
Unless there's any major bugs (e.g. disconnecting or client breaks from basic use -- and again: skills not working, or such things are not really relevant to the PR - they'll require their own concerted effort to update/implement), I'd say it's pretty much ready to go.


Client/N3Base/N3ShapeMgr.h
```diff
@@ -37,138 +37,175 @@ class CN3ShapeMgr : public CN3BaseFileAccess
...
+	float		GetHeightNearstPos(const __Vector3& vPos, float fDist, __Vector3* pvNormal = nullptr);
+
+	// 가장 가까운 높이을 돌려준다. 없으면 -FLT_MAX 을 돌려준다.
+	float		GetHeightNearstPos(const __Vector3& vPos, __Vector3* pvNormal = nullptr);
+
+	// 현재 지점에서 제일 높은 값을 돌려준다. 없으면 -FLT_MAX 을 돌려준다.
+	float		GetHeight(float fX, float fZ, __Vector3* pvNormal = nullptr);
+
+	// 벡터 사이에 걸친 셀포인터 돌려준다..
+	int			SubCellPathThru(const __Vector3& vFrom, const __Vector3& vAt, int iMaxSubCell, __CellSub** ppSubCells);
```
kenner2, line 196: Probably wouldn't hurt to hit this file with the formatter.  There's a lot of weird spacing

Server/AIServer/Compress.cpp
```diff
@@ -0,0 +1,360 @@
...
+{
+	short l_len;
+
+	if (addList == nullptr)
+		addList = new CObList;
+
+	char* dataPtr;
+	if (nLen == 0)
+		l_len = static_cast<short>(strlen(add_data));
+	else
```
kenner2, line 43: nit: can avoid the else/cast by initializing l_len to 0

kenner2: nevermind I misread

Server/AIServer/Compress.cpp
```diff
@@ -0,0 +1,360 @@
...
+{
+	int iStatus;
+
+	if (OrgDataLength == 0)
+		return true;
+
+	// COMPRESS THE FILE
+	iStatus = implode(ReadBuffer, WriteBuffer, pScratchPad, this, &DataType, &DictSize);
+	if (iStatus != 0)
+		return false;
```
kenner2, line 230: Non-issue as is, but this one's a little weird to me.  implode can return 0 (no error), 1 (CMP_INVALID_DICTSIZE), or 2 (CMP_INVALID_MODE).  Feels like we should either return a bool or just return the iStatus so we know what error code we got.  

That said, all calls to Compress() discard the return value so there's no real functional impact.

Server/AIServer/Compress.cpp
```diff
@@ -0,0 +1,360 @@
...
+	POSITION pos;
+	CCompressData* pData;
+
+	if (addList != nullptr)
+	{
+		if (addList->GetCount() > 0)
+		{
+			for (pos = addList->GetHeadPosition(); pos != nullptr; )
+			{
+				pData = (CCompressData*) addList->GetNext(pos);
```
kenner2, line 246: Should this be using addList->GetAt()?  Unless I'm missing something it looks like the list head would go undeleted.  Unless `delete addList` deletes the head, and `addList->GetHeadPosition()` is kind of redundant

twostars: Yeah I don't really know. I think it's better to just drop pkware and this awful mess they created for it, and simplify it with LZF. Then we don't have to worry about whatever nonsense they did here.

Server/AIServer/Compress.cpp
twostars: Yeah it's... assuming it's appending string data here to compress if no length is supplied. It's really wonky.

PR #253, opened 2025-06-28
CN3UITooltip appearance

koraykarakus:
## Pull request type

- [x] Bugfix

tooltip text is not filled with parameters it shows like type %s. Tooltip size is not correct, text is not visible.

## What is the new behaviour?

text filled with strings, tooltip size corrected.

visual bug preventing user to read what command does.

## Demo

tooltip

## Checklist


UTengine:
Can you try giving yourself like 3-4 buffs and see if tooltip is still within window for each of them?

koraykarakus:
buff
buff tip seems ok

UTengine:
Weird never had issues with tooltips being cut off.
But then again I didn't use openko that much in the past. Anout the %s i think it should be because of sprintf buff then copying it to cmdedit or w.e.
So when you treat it as a std string and pass it on you get issues like %s. You could do the same with the newer std functions though.

UTengine:
I just don't understand the code. What I did back in the days was just tie the texts to every command of you type them or not. And then the cmd would just be bound to the chatcmd. It would just execute like when you have someone selected as a player and select PM it would open PM box or trade box etc. I'm really confused but maybe i'm overthinking it. Like most commands are already there and are executed by buttons too. Idk fk it

koraykarakus:
those works with opening cmdedit, at current state it only works for PM, but it should work with ban, mute etc. when you double clicked on forbiduser cmdedit should open and expect an input of username.

twostars:
Officially the tooltips wrap at 500px. I recommend checking against the official 1.298 client when testing such behaviour.

koraykarakus:
where is official client, do you have any link without virus ?

twostars:
I'm really hesitant to do anything with this PR. We're doing way too much here. We're trying to rework CN3UITooltip's display logic, and we're trying to implement various related and unrelated CUICmdList logic.

I think this PR effectively needs to be split up so it can focus on its issues separately. It's difficult to address everything in this PR because there's just so much unrelated things going on here.

For one, CN3UITooltip's logic - which you even say still doesn't match official's - is a problem.
Beyond that, between the random hardcoded resource IDs that are just excluded and such, it's... just a lot. I don't really know where to even begin reviewing this.

I suggest taking this branch and splitting it up, focusing on CN3UITooltip and getting that behaving like official, and then the various CUICmdList tweaks. It's a lot easier to just discuss each individual change that way.

twostars:
Closing this as it appears to be abandoned.

PR #254, opened 2025-06-28
weight text update while trading with NPC

koraykarakus:
## Pull request type

- [x] Bugfix
- [x] Feature

weight text is not used although it presents on uif file. User cannot see weight change on the state of trading with NPC.

## What is the new behaviour?

weight is updated after trade, and text is visible on ui.

it was missing core feature

## Checklist


twostars:
The weight is updated with `WIZ_WEIGHT_CHANGE`. As such, it should just update corresponding UIs from there, instead of manually sprinkling in updates everywhere.

This way you can reuse the existing formatted strings for it as well.

koraykarakus:
made changes accordingly on commit 2.

PR #256, opened 2025-06-28
level guide

koraykarakus:
## Pull request type

- [x] Bugfix
- [x] Feature

Level guide does not open.

## What is the new behaviour?

Level guide works ok.

one of the missing core feature.

## Demo

levelguide

## Checklist


Client/WarFare/WarFare.vcxproj.user
```diff
@@ -1,7 +1,7 @@
 ﻿<?xml version="1.0" encoding="utf-8"?>
 <Project ToolsVersion="14.0" xmlns="http://schemas.microsoft.com/developer/msbuild/2003">
   <PropertyGroup Condition="'$(Configuration)|$(Platform)'=='Debug|Win32'">
-    <LocalDebuggerWorkingDirectory>..\Data\</LocalDebuggerWorkingDirectory>
```
twostars: This shouldn't be in the PR

Client/WarFare/UILevelGuide.cpp
```diff
@@ -0,0 +1,385 @@
...
+	m_pTextGuide0 = nullptr;
+	m_pTextTitle2 = nullptr;
+	m_pTextTitle1 = nullptr;
+	m_pTextTitle0 = nullptr;
+	m_pTextLevel = nullptr;
+	m_pButtonUp = nullptr;
+	m_pButtonDown = nullptr;
+	m_pButtonCancel = nullptr;
+
+	m_iUserLevel = CGameBase::s_pPlayer->m_InfoBase.iLevel;
```
twostars: This shouldn't be stored. As-is, it's it's only ever using whatever it's set to when the UI actually loads, which is most likely before the WIZ_MYINFO packet.

Just check the level directly and you don't have to worry about that.

Client/WarFare/UILevelGuide.cpp
```diff
@@ -0,0 +1,385 @@
...
+		if (m_pEditLevel && m_pEditLevel->HaveFocus())
+			m_pEditLevel->KillFocus();
+	}
+}
+
+bool CUILevelGuide::ReceiveMessage(CN3UIBase* pSender, uint32_t dwMsg)
+{
+	if (dwMsg == UIMSG_BUTTON_CLICK)
+	{
+		if (pSender == (CN3UIBase*) m_pButtonCancel)
```
twostars: All of these casts can go. Just add the include for <N3Base/N3UIButton.h> in this source file and you'll be fine.

Client/WarFare/UILevelGuide.cpp
```diff
@@ -0,0 +1,385 @@
...
+		{
+			this->SetVisible(false);
```
twostars: "this->" can be stripped

Client/WarFare/UILevelGuide.cpp
```diff
@@ -0,0 +1,385 @@
...
+			return false;
+		}
+	}
+	else if (dwMsg == UIMSG_SCROLLBAR_POS)
+	{
+		int iCurLinePos = 1;
+		CN3UIScrollBar* pScroll = nullptr;
+		CN3UIString* pText = nullptr;
+
+		if (pSender == (CN3UIBase*) m_pScrollGuide0)
```
twostars: These casts can also be removed. Again: add the control's include in this source file and you'll be fine.

Client/WarFare/UILevelGuide.cpp
```diff
@@ -0,0 +1,385 @@
...
+
+bool CUILevelGuide::OnKeyPress(int iKey)
+{
+	switch (iKey)
+	{
+		case DIK_ESCAPE:
+			this->SetVisible(false);
+			return true;
+		case DIK_TAB:
+			if (m_pEditLevel)
```
twostars: This should read:
```cpp
if (m_pEditLevel != nullptr)
```

Client/WarFare/UILevelGuide.cpp
```diff
@@ -0,0 +1,385 @@
...
+			{
+				m_pEditLevel->SetFocus();
+
+				std::string strText = m_pEditLevel->GetString();
```
twostars: There is no reason for this to copy the string.
This should be a const reference, i.e.
```cpp
const std::string& strText = m_pEditLevel->GetString();
```

Client/WarFare/UILevelGuide.cpp
```diff
@@ -0,0 +1,385 @@
...
+
+				size_t stlastPos = strText.length();
+
+				//BUG: set position to end, but when user keypress it resets
+				if (stlastPos <= UINT_MAX) //avoid overflow
```
twostars: I would argue that this is pointless - just cap the control's limit if it's a problem.

Client/WarFare/UILevelGuide.cpp
```diff
@@ -0,0 +1,385 @@
...
+					m_pEditLevel->SetCaretPos(static_cast<UINT>(stlastPos));
+
+			}
+			return true;
+		case DIK_RETURN:
+			if (m_pEditLevel && IsNumber(m_pEditLevel->GetString()))
```
twostars: Again, this should read:
```cpp
if (m_pEditLevel != nullptr...)
```

Client/WarFare/UILevelGuide.cpp
```diff
@@ -0,0 +1,385 @@
...
+				m_iSearchLevel = std::stoi(m_pEditLevel->GetString());
```
twostars: std::stoi() can throw exceptions. Unless you're willing to handle those, just stick to atoi().

Client/WarFare/UILevelGuide.h
```diff
@@ -0,0 +1,57 @@
...
+	CN3UIString* m_pTextGuide0;
+	CN3UIString* m_pTextTitle2;
+	CN3UIString* m_pTextTitle1;
+	CN3UIString* m_pTextTitle0;
+	CN3UIString* m_pTextLevel;
+	CN3UIButton* m_pButtonUp;
+	CN3UIButton* m_pButtonDown;
+	CN3UIButton* m_pButtonCancel;
+
+	const int m_iMaxLevel = 80;
```
twostars: Is this intended to be the level cap? Use the existing constant for that; you don't need to define it here.

Client/WarFare/UILevelGuide.h
```diff
@@ -0,0 +1,57 @@
...
+	const int m_iRangeSearch = 5; //ex: searches 5 levels up and down
```
twostars: This should just be `static constexpr int`. It doesn't need to be a member, since it's just an arbitrary constant related to this class.

And then following on from that, the `m_` denotes that it's a member variable, so you'd change that. I would just restructure it to like, MAX_LEVEL_RANGE or something.

Client/WarFare/UILevelGuide.h
```diff
@@ -0,0 +1,57 @@
...
+	static constexpr int m_iContentCount = 20; //number of contents, equal to m_saQuestTitle,m_saQuestText size
```
twostars: As above: since it's not actually a member variable, it shouldn't be prefixed with `m_` (which is the Hungarian notation denoting that it is a member variable. Which this is not.).

Client/WarFare/UILevelGuide.h
```diff
@@ -0,0 +1,57 @@
...
+	int m_iUserLevel;
+	int m_iSearchLevel;
+	int m_iCurrentPage;
+	std::string m_saQuestTitle[m_iContentCount];
+	std::string m_saQuestText[m_iContentCount];
+	static bool IsNumber(const std::string& s);
+	void SetTopLine(CN3UIScrollBar* pScroll, CN3UIString* pTextGuide);
+public:
+	virtual bool	Load(HANDLE hFile);
```
twostars: These aren't defining new virtual methods, they're defining overrides.
As such, the `virtual` portion should be dropped and `override` should be added to the end - this forces the compiler to verify that it matches an existing defined virtual method with the same signature, as opposed to unintentionally creating a new one that nothing will ever really use if it ever happens to be wrong in some way.

Client/WarFare/UILevelGuide.h
```diff
@@ -0,0 +1,57 @@
...
+	virtual void	Release();
+	CUILevelGuide();
+	virtual ~CUILevelGuide();
+	void SetVisible(bool bVisible);
```
twostars: This is a virtual method (but we should append `override`), so:
```cpp
void SetVisible(bool bVisible) override;
```

Client/WarFare/UILevelGuide.h
```diff
@@ -0,0 +1,57 @@
...
+	bool ReceiveMessage(CN3UIBase* pSender, uint32_t dwMsg);
```
twostars: This is a virtual method (but we should append `override`).

Client/WarFare/UILevelGuide.h
```diff
@@ -0,0 +1,57 @@
...
+	bool OnKeyPress(int iKey);
```
twostars: This is a virtual method (but we should append `override`).

Client/WarFare/UIStateBar.cpp
```diff
@@ -588,8 +589,13 @@ bool CUIStateBar::ReceiveMessage(CN3UIBase* pSender, uint32_t dwMsg)
 		}
 		else if (pSender == (CN3UIBase*)m_pBtn_Quest)
 		{
-			//TODO: Show Leveling Guide UI (Quest Helper)
-			return false;
+			if (CGameProcedure::s_pProcMain->m_pUILevelGuide != nullptr)
```
twostars, line 591: It would be convenient to just fetch it to a var and use it for all 3 uses, instead of being forced to do the full lookup for it all every time.

e.g.
```cpp
CUILevelGuide* pUILevelGuide = CGameProcedure::s_pProcMain->m_pUILevelGuide;
if (pUILevelGuide != nullptr)
{
...
}
```

Client/WarFare/UILevelGuide.cpp
```diff
@@ -0,0 +1,385 @@
...
+		m_saQuestTitle[i].clear();
+
+	for (int i = 0; i < 10; i++)
+		m_saQuestText[i].clear();
+
+}
+
+bool CUILevelGuide::Load(HANDLE hFile)
+{
+	if (CN3UIBase::Load(hFile) == false) return false;
```
twostars:
```cpp
	if (!CN3UIBase::Load(hFile))
            return false;
```

Client/WarFare/UILevelGuide.cpp
twostars: After going over it, I see that you don't even use it. So it should be removed altogether.

Client/WarFare/UILevelGuide.cpp
```diff
@@ -0,0 +1,385 @@
...
+
+
+
+		iStartIndex++;
+	}
+
+	if (m_pTextPage != nullptr)
+	{
+		std::string strCurrentPage = std::to_string(m_iCurrentPage);
+		m_pTextPage->SetString(strCurrentPage);
```
twostars: `SetStringAsInt()` also exists and is typically how this is done, though some of them still do it manually.

Client/WarFare/UILevelGuide.cpp
```diff
@@ -0,0 +1,385 @@
...
+	}
+
+
+
+}
+
+bool CUILevelGuide::IsNumber(const std::string& s)
```
twostars: I don't really like this being in `CUILevelGuide`. It's not really logic specific to this class whatsoever
Based on the existing usage this fits better in CN3UIEdit, buuuuuuut I really don't think it's altogether that necessary.

I think I'd simplify this by just going back to just atoi()'ing it -- you can keep it checking if the string is set, sure, but what would make more sense is filtering it on the control so it only allowed numerical input, rather than it allowing all input and hoping for the best.

Client/WarFare/UILevelGuide.cpp
```diff
@@ -0,0 +1,385 @@
...
+		else if (pSender == (CN3UIBase*) m_pButtonDown)
+		{
+			m_iCurrentPage--;
+			LoadContent();
+			return true;
+		}
+		else if (pSender == (CN3UIBase*) m_pButtonCheck) //search button
+		{
+
+			std::string strSearchLevel;
```
twostars: Realistically I would restructure this as explained previously (drop IsNumber(), filter numerical input only in the control instead, use atoi() and not bother dealing with exceptions), but that aside, this should be a const reference (it would need to be reshuffled slightly).

Client/WarFare/GameBase.h
```diff
@@ -21,6 +21,7 @@ class CGameBase : public CN3Base
 	static CN3TableBase<__TABLE_FX>					s_pTbl_FXSource;		// FX소스 정보 테이블..
 	static CN3TableBase<__TABLE_QUEST_MENU>			s_pTbl_QuestMenu;		// 퀘스트 선택 메뉴
 	static CN3TableBase<__TABLE_QUEST_TALK>			s_pTbl_QuestTalk;		// 퀘스트 지문
+	static CN3TableBase<__TABLE_QUEST_CONTENT>		s_pTbl_Quest_Content;	// Quest Content Table
```
twostars: I think all 3 of these names should be named consistently. They should either be Quest_Menu, Quest_Talk, Quest_Content etc, or QuestMenu, QuestTalk, QuestContent.
Just because they're already named like this, I would suggest just renaming this to s_pTbl_QuestContent for consistency.

twostars (changes requested):
(required comment)

Client/WarFare/UILevelGuide.h
koraykarakus: Defined those as member because code can be expanded for different needs in the future, for example one can change range for his/her needs or modify contentcount etc. about maxlevel I remember getting errors when I try to include related cpp file in which it is defined. But will turn back to fixing when I have extra time.

Client/WarFare/UIStateBar.cpp
koraykarakus: this need a new togglefunction inside gameprocmain.cpp, because opening this should close other UI such as skill tree, inventory, cmd list, so i changed that in my last commit.

Client/WarFare/UILevelGuide.cpp
koraykarakus: tab feature is removed, it is not official.

Client/WarFare/UILevelGuide.h
koraykarakus: only constant i could find is placed inside server Define.h ( MAX_LEVEL )

twostars: Hm, OK. Well when we merge in the server code PR, this will be better shared between the 2 of them, though in this case this needs moving across into the header too.

twostars:
As of 86e31fa513cb278555b2344579c9c3759fa64b77, MAX_LEVEL should be available through <shared/globals.h> (which is included by GameDef.h already).

Client/WarFare/UILevelGuide.h
```diff
@@ -33,18 +33,18 @@ class CUILevelGuide : public CN3UIBase
 	CN3UIButton* m_pButtonCancel;
 
 	const int m_iMaxLevel = 80;
-	const int m_iRangeSearch = 5; //ex: searches 5 levels up and down
-	static constexpr int m_iContentCount = 20; //number of contents, equal to m_saQuestTitle,m_saQuestText size
+	static constexpr int iRangeSearch = 5; //ex: searches 5 levels up and down
```
twostars: The problem now is they look like local vars. Either drop the Hungarian notation (so it's just "RangeSearch") or throw it in uppercase (e.g. "RANGE_SEARCH") and it'll stand out more that way. But switch it around so it's 'search range' - it makes more sense that way.

Client/WarFare/GameProcMain.cpp
```diff
@@ -4732,6 +4747,9 @@ bool CGameProcMain::CommandToggleUIInventory()
 	{
 		bNeedOpen = true;
 		if(m_pUISkillTreeDlg->IsVisible()) m_pUISkillTreeDlg->Close();
+
+		if (m_pUILevelGuide->IsVisible())
```
twostars: All of these visibility updates are unnecessary. Add the style UISTYLE_POS_RIGHT and it'll be mutually exclusive with them (this is the official behaviour).

twostars:
Okay, after going through and tidying the rest up, the only remaining issue I have is the hard internal list limit.
The problem is, this doesn't really exist for any particular reason at all.

You do it for storage only, but there's not even a need to store these - it can just be figured out on-demand via the search and page updates. All it needs to know is which page we're on, so it can figure out which results to start from.

Fix that and I think we're fairly good to merge it in.

koraykarakus:
correct folder and opened those with correct order (tested now). But I have edited database a lot, probably it is caused by those changes I made on my local project. Need to reinstall DB, but don't have spare time nowadays, I ll take a loot at in the future.
ebenezer log similar to this

before item
before main
before 1
before 2
before 3
before 4
before 5
before 8
before Coeffi
before Level

which does not mean anything to me at all

kenner2:
"before Level" is logged right before the LEVEL_UP table is loaded.  If you're not getting "before All Kinghts" in the log, then something is causing the application to crash while loading the LEVEL_UP table

Client/WarFare/UILevelGuide.cpp
```diff
@@ -132,63 +150,77 @@ void CUILevelGuide::LoadContent()
...
+	auto& questDataMap = CGameBase::s_pTbl_Help.GetMap();
+	for (const auto& [_, questData] : questDataMap)
 	{
-		__TABLE_QUEST_CONTENT* pQuestContent = CGameBase::s_pTbl_QuestContent.GetIndexedData(i);
-		if (pQuestContent == nullptr)
-			continue;
+		if (iSearchLevel == questData.iMinLevel
+			&& iSearchLevel == questData.iMaxLevel)
+		{
+			if (questData.iReqClass == CLASS_REPRESENT_UNKNOWN
```
kenner2: nested ifs can be combined (same for the else if)

kenner2: actually don't know why we would want to check exact equal in the if and range check in the else if when the nested ifs are the same.

PR #257, opened 2025-06-28
better solution to weight bug, which works in both clients

koraykarakus:
## Pull request type

- [x] Bugfix

weight works fine in compiled client but it does not work on binary client. Hence, instead of static casting int, weight limit set as INT16_MAX. Probably, officially it was in that way.

## What is the new behaviour?

Max weight set as INT16_MAX on server side works in both clients.

Old fix was not working on original exe.

## Checklist

### DIFF
--- a/Client/WarFare/GameProcMain.cpp
+++ b/Client/WarFare/GameProcMain.cpp
@@ -3299,7 +3299,7 @@ bool CGameProcMain::MsgRecv_ItemMove(Packet& pkt)
 	{
 		pInfoExt->iAttack = pkt.read<int16_t>();
 		pInfoExt->iGuard =	pkt.read<int16_t>();
-		pInfoExt->iWeightMax = static_cast<int>(pkt.read<uint16_t>());
+		pInfoExt->iWeightMax = pkt.read<int16_t>();
 		
 		pInfoBase->iHPMax = pkt.read<int16_t>();
 		pInfoExt->iMSPMax = pkt.read<int16_t>();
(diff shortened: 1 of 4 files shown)


kenner2:
> compiled client but it does not work on binary client

What does that mean?  It works when running inside of an IDE but not the built executable?   IDE's typically build and run an executable out of a temp folder.

The big difference between a static cast and a c-style cast is that static casts are checked at compile time for safety, where c-style casts won't cause an error until their code is hit at runtime.  Seems like something else might be off

twostars:
> > compiled client but it does not work on binary client
> 
> What does that mean? It works when running inside of an IDE but not the built executable? IDE's typically build and run an executable out of a temp folder.
> 
> The big difference between a static cast and a c-style cast is that static casts are checked at compile time for safety, where c-style casts won't cause an error until their code is hit at runtime. Seems like something else might be off

He means the official 1.298 client EXE here. Essentially he reverted the client stuff to how it was before his previous change, and capped it in the server.

twostars:
Okay so the only real problem with this now is that it's left to be a server change. Which wouldn't *normally* be an issue, but all of this server code is getting replaced shortly anyway once #248 is merged in.

image

We can look at restoring it after it's merged in.

koraykarakus:
I refer leaked orginal client as binary. Back in the day probably they used max weight around INT16_MAX. I remember something like that existed but not sure.
At current state compiled exe from openko project works, but original one does not. Maybe it can be cast on server side for increased weight limit. 

Sent from Yahoo Mail for iPhone


On Pazar, Haziran 29, 2025, 3:08 ÖÖ, Ken R. ***@***.***> wrote:

kenner2 left a comment (Open-KO/KnightOnline#257)

compiled client but it does not work on binary client


What does that mean? It works when running inside of an IDE but not the built executable? IDE's typically build and run an executable out of a temp folder.

The big difference between a static cast and a c-style cast is that static casts are checked at compile time for safety, where c-style casts won't cause an error until their code is hit at runtime. Seems like something else might be off

PR #282, opened 2025-07-04
gold formatting for transactions

koraykarakus:
## Pull request type

- [x] Bugfix

formating not used in transactions UI, although it is used in original version.

## What is the new behaviour?

gold is formatted.

official feature

## Checklist

Note: after last changes, AI does not work properly, some NPCs are loaded but most not, mobs not loaded as well. Transactions don't get sell item packet from server ( selling item does not grands you gold ) , Seems some issues with ebenezer as well, so I could only partially test this PR.


twostars (changes requested):
Please pre-format these (storing them in a local variable) before passing them to SetString(), since this uses pass-by-reference semantics.

This also then allows you to avoid the case where you're having to format it twice, once for each control in `CUITransactionDlg::ReceiveResultTradeFromServer()`.

You also don't need to check the pInfoExt pointers here. Realistically these should be references in general since their intent in making them pointers here is to simplify access (reducing having to specify the entire thing - `CGameBase::s_pPlayer->m_InfoExt.`), but they don't need to be checked.

PR #294, opened 2025-07-08
Automate dependency builds

twostars:
Ideally dependencies should only be built once, which is why we were initially distributing precompiled libraries.

As we found, doing this is quite fragile - even minor MSVC compiler versions can break this, so we swapped to requiring the user to one-time build these themselves.

While in practice it was just a matter of running a single script, this still leads to confusion, as well as extra documentation specifically highlighting this. Additionally, come time for these submodules to be updated, users will have to simply know to update things again (i.e. just run the script).

So, instead, the idea is to essentially attempt to update & build on-demand with wrapper projects.

These wrapper projects simply call build_scripts/build_dependency.cmd to:
1. Find & compare the last build state (which we store per dependency, per build configuration and exclude in .gitignore). This is just a simple text file containing a very simple key (commit hash;VC toolset version). Both of these components are needed for verifying if it should rebuild, since it should be updated: - When the submodule itself is updated (since it won't do this automatically otherwise) - When the compiler updates (to be safe, since it can break)

2. If the build state has not changed, its build can stop here.

3. If it HAS changed, it will initialize and update the associated submodule. Additionally, if it has a project to build (e.g. zlib), it will rebuild that project. Dependencies like dx9sdk or ko-client-assets have no associated project to build -- we just need to keep the submodule updated.

And that, essentially is the crux of it. Only changes should affect it. There is technically a little overlap here; it *can* erroneously update & initialize the submodule when simply switching build config, since it's affected by the shared build state, but this likely isn't worth the effort of separating since it's a rare case and won't do much of anything to the submodule itself anyway -- it'll typically be a very fast no-op.

In effect, this means we can just build the relevant project and have any dependencies needed updated and built for us -- but only as needed, NOT forced to rebuild every time we invoke a full rebuild on the project/solution.

Why wrapper projects instead of pre-build steps? Shared dependencies.
If multiple projects require zlib (which they do, in All.sln), you don't want the script running on both.
So we take advantage of MSVC's existing build dependency system to ensure they know they both depend on zlib, so that project should build before either of them.

Though it's not really needed anymore, ThirdParty.sln still exists for manually building these projects. I think it's probably helpful to see all such projects in one place, especially for future migration.

### DIFF
--- a/build_scripts/env_find_git.cmd
+++ b/build_scripts/env_find_git.cmd
@@ -2,8 +2,8 @@
 SETLOCAL
 
 SET "GitDir="
-WHERE /q git
-IF NOT ERRORLEVEL 1 (
+where /q git
+IF ERRORLEVEL 0 (
 	FOR /f "delims=" %%G IN ('where git') DO (
 		SET "GitPath=%%G"
 		GOTO found_git


build_scripts/env_find_git.cmd
```diff
@@ -0,0 +1,38 @@
...
+SET "GitDir="
+WHERE /q git
+IF NOT ERRORLEVEL 1 (
+	FOR /f "delims=" %%G IN ('where git') DO (
+		SET "GitPath=%%G"
+		GOTO found_git
+	)
+)
+
+SET GithubDir=%LOCALAPPDATA%\GitHubDesktop
```
kenner2, line 13: Is this going to cause an issue for people not using Github's CLI program?

twostars: This is fallback. It prefers using git.exe from %PATH%, then looks to see if Github Desktop's installed.
If it can't find either, it fails.

PR #296, opened 2025-07-08
right click item from inventory to hotkey

koraykarakus:
## Pull request type

- [x] Feature

cannot move consumables from inventory to skill bar

## What is the new behaviour?

can move items from inventory to skillbar with right click mouse button.

core feature

## Checklist

## Summary by Sourcery

Enable quick assignment of consumable inventory items to skill hotkey bar via right-click.

New Features:
- Allow right-clicking on usable (consumable) inventory items to map them to the first empty hotkey slot.
- Introduce CUIHotKeyDlg::SetReceiveSelectedItem to create and place skill-like icons for inventory items on the hotkey bar.


Client/WarFare/UIHotKeyDlg.cpp
```diff
@@ -933,6 +933,55 @@ bool CUIHotKeyDlg::ReceiveIconDrop(__IconItemSkill* spItem, POINT ptCur)
...
+	if (pUSkill == nullptr) return false;
+	if (pUSkill->dwID < UIITEM_TYPE_SONGPYUN_ID_MIN) return false;
+
+	spSkill = new __IconItemSkill();
+	spSkill->pSkill = pUSkill;
+
+	// Create the icon name
+	std::vector<char> buffer(256, NULL);
+	sprintf(&buffer[0], "UI\\skillicon_%.2d_%d.dxt", spItem->pItemBasic->dwEffectID1 % 100, spItem->pItemBasic->dwEffectID1 / 100);
+	spSkill->szIconFN = &buffer[0];
```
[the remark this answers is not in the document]
twostars: Not necessarily accurate.

I would instead recommend changing this:
```cpp
std::vector<char> buffer(256, NULL);
```
To a standard array, i.e.:
```cpp
char buffer[256] = {};
```

It has no need to be on the heap.

Then for both the sprintf() and the assignment, you'd use `buffer` directly, instead of `&buffer[0]`;

Client/WarFare/UIHotKeyDlg.cpp
```diff
@@ -933,6 +933,55 @@ bool CUIHotKeyDlg::ReceiveIconDrop(__IconItemSkill* spItem, POINT ptCur)
...
+
+	__TABLE_UPC_SKILL* pUSkill = CGameBase::s_pTbl_Skill.Find(spItem->pItemBasic->dwEffectID1);
+	if (pUSkill == nullptr) return false;
+	if (pUSkill->dwID < UIITEM_TYPE_SONGPYUN_ID_MIN) return false;
+
+	spSkill = new __IconItemSkill();
+	spSkill->pSkill = pUSkill;
+
+	// Create the icon name
+	std::vector<char> buffer(256, NULL);
```
twostars: I would instead recommend changing this:
```cpp
std::vector<char> buffer(256, NULL);
```
To a standard array, i.e.:
```cpp
char buffer[256] = {};
```

It has no need to be on the heap.

Then for both the sprintf() and the assignment, you'd use `buffer` directly, instead of `&buffer[0]`;

Client/WarFare/UIHotKeyDlg.cpp
```diff
@@ -933,6 +933,55 @@ bool CUIHotKeyDlg::ReceiveIconDrop(__IconItemSkill* spItem, POINT ptCur)
...
+	sprintf(&buffer[0], "UI\\skillicon_%.2d_%d.dxt", spItem->pItemBasic->dwEffectID1 % 100, spItem->pItemBasic->dwEffectID1 / 100);
+	spSkill->szIconFN = &buffer[0];
+
+	// load icon
+	spSkill->pUIIcon = new CN3UIIcon;
+	spSkill->pUIIcon->Init(this);
+	spSkill->pUIIcon->SetTex(spSkill->szIconFN);
+	spSkill->pUIIcon->SetUVRect(0, 0, 1.0f, 1.0f);
+	spSkill->pUIIcon->SetUIType(UI_TYPE_ICON);
+	spSkill->pUIIcon->SetStyle(UISTYLE_ICON_SKILL);
```
twostars: This is useless. It's set again immediately after.

Client/WarFare/UIHotKeyDlg.cpp
```diff
@@ -933,6 +933,55 @@ bool CUIHotKeyDlg::ReceiveIconDrop(__IconItemSkill* spItem, POINT ptCur)
 	return false;
 }
 
+bool CUIHotKeyDlg::SetReceiveSelectedItem(int iIndex)
+{
+	if (CN3UIWndBase::m_sSelectedIconInfo.UIWndSelect.UIWnd != UIWND_INVENTORY)
+		return false;
+	
+	__IconItemSkill* spSkill, * spItem;
```
twostars: I would just not bother pre-declaring them here, but that space shouldn't exist after the second `*`.

Client/WarFare/UIHotKeyDlg.cpp
```diff
@@ -933,6 +933,55 @@ bool CUIHotKeyDlg::ReceiveIconDrop(__IconItemSkill* spItem, POINT ptCur)
...
+	CN3UIArea* pArea = nullptr;
+	pArea = CN3UIWndBase::GetChildAreaByiOrder(UI_AREA_TYPE_SKILL_HOTKEY, iIndex);
+	
+	if (pArea != nullptr)
+	{
+		spSkill->pUIIcon->SetRegion(pArea->GetRegion());
+		spSkill->pUIIcon->SetMoveRect(pArea->GetRegion());
+	}
+	
+	m_pMyHotkey[m_iCurPage][iIndex] = spSkill;
```
twostars, line 984: This assumes that this isn't already set. Since this doesn't expect to replace, I would check this simply pre-check this before allocation.

Client/WarFare/UIInventory.cpp
```diff
@@ -1484,16 +1485,37 @@ bool CUIInventory::ReceiveMessage(CN3UIBase* pSender, uint32_t dwMsg)
...
+					if (bSlot) //player clicked on equiped items
+					{
+						CN3UIWndBase::m_sSelectedIconInfo.UIWndSelect.UIWndDistrict = UIWND_DISTRICT_INVENTORY_SLOT;
+					}
+					else //player clicked on item which is not equiped
+					{
+						CN3UIWndBase::m_sSelectedIconInfo.UIWndSelect.UIWndDistrict = UIWND_DISTRICT_INVENTORY_INV;
+						
+						//right click on item to hotkey (skillbar)
+						if (spItem->pItemBasic->byClass == 97) //usable items
```
twostars: Magic numbers are bad. You will find this to be:
```cpp
						ITEM_CLASS_POTION = 97, // 물약
```

On a related note, ensure we include a space after the '//' in a comment, and try not to position them horizontally like this. This one's unnecessary because we have a cosntant, but the ones on the if () else () statements immediately above need shifting above.

Client/WarFare/UIInventory.cpp
```diff
@@ -1484,16 +1485,37 @@ bool CUIInventory::ReceiveMessage(CN3UIBase* pSender, uint32_t dwMsg)
...
+						{
+							CUIHotKeyDlg* pDlg = CGameProcedure::s_pProcMain->m_pUIHotKeyDlg;
+							int iIndex;
+							if (pDlg->GetEmptySlotIndex(iIndex))
+							{
+								bool bResult = false;
```
twostars: This doesn't really need to exist at all.

Client/WarFare/UIInventory.cpp
```diff
@@ -1484,16 +1485,37 @@ bool CUIInventory::ReceiveMessage(CN3UIBase* pSender, uint32_t dwMsg)
...
+								CN3UIWndBase::m_sSkillSelectInfo.UIWnd = UIWND_INVENTORY;
```
twostars: This... is also pointless. You explicitly add this to explicitly check it in the method you added.
Since it's only for this case, you can just expect it to be correct and for any relevant info under "m_sSkillSelectInfo" (i.e. the icon) you can just pass in directly.

Client/WarFare/UIHotKeyDlg.cpp
koraykarakus: commenting out those lines results in empty icons, it works but now picture not seen on skill bar.

Client/WarFare/UIInventory.cpp
koraykarakus: there is a confusion here. when you successfully put an item on to hotkey dlg no sound is played. However, when you clicked on item which cannot be put into hotkey (for example: abbys gem), item sound is played ( sound_etc ). If I remove bool and return true in all cases, than it never plays the sound on fail. 

By the way that sound level is too litle, I barely can hear it on max level.

Client/WarFare/UIInventory.cpp
koraykarakus: yes just saw it yesterday, that is fixed on commit. Maybe we should change its name though, ITEM_CLASS_POTION also covers some items which are not potions ( like transformation gem, catapult items etc. )

Client/WarFare/UIInventory.cpp
twostars: I just meant that it doesn't need to store a bool here at all. It can just act based on the method's result directly.

Client/WarFare/UIHotKeyDlg.cpp
twostars: "those" lines? You only need to remove the first one because it's set twice is my point.

Client/WarFare/UIHotKeyDlg.cpp
```diff
@@ -933,6 +933,55 @@ bool CUIHotKeyDlg::ReceiveIconDrop(__IconItemSkill* spItem, POINT ptCur)
...
+	__TABLE_UPC_SKILL* pUSkill = CGameBase::s_pTbl_Skill.Find(spItem->pItemBasic->dwEffectID1);
+	if (pUSkill == nullptr) return false;
+	if (pUSkill->dwID < UIITEM_TYPE_SONGPYUN_ID_MIN) return false;
+
+	spSkill = new __IconItemSkill();
+	spSkill->pSkill = pUSkill;
+
+	// Create the icon name
+	char buffer[256] = {};
+	sprintf(&buffer[0], "UI\\skillicon_%.2d_%d.dxt", spItem->pItemBasic->dwEffectID1 % 100, spItem->pItemBasic->dwEffectID1 / 100);
```
twostars: &buffer[0] should just be buffer, this line and the line below it

Client/WarFare/UIHotKeyDlg.cpp
```diff
@@ -976,7 +975,10 @@ bool CUIHotKeyDlg::SetReceiveSelectedItem(int iIndex)
 		spSkill->pUIIcon->SetMoveRect(pArea->GetRegion());
 	}
 	
-	m_pMyHotkey[m_iCurPage][iIndex] = spSkill;
+	if (m_pMyHotkey[m_iCurPage][iIndex] == nullptr)
```
twostars: I think you misunderstood. You just swapped the potential memory leak from one side to the other with this check; now spSkill (and pUIIcon) will leak, rather than the original icon that you were blindly replacing.

As I suggested, add the preventative check to the start - there's no reason for it to be set in the first place, so you can just pre-check it at the start of the method and early fail.

koraykarakus:
some of the change requests does not seem on github, review please

Client/WarFare/UIInventory.cpp
```diff
@@ -1494,16 +1495,34 @@ bool CUIInventory::ReceiveMessage(CN3UIBase* pSender, uint32_t dwMsg)
...
+					if (bSlot) //player clicked on equiped items
+					{
+						CN3UIWndBase::m_sSelectedIconInfo.UIWndSelect.UIWndDistrict = UIWND_DISTRICT_INVENTORY_SLOT;
+					}
+					else //player clicked on item which is not equiped
+					{
+						CN3UIWndBase::m_sSelectedIconInfo.UIWndSelect.UIWndDistrict = UIWND_DISTRICT_INVENTORY_INV;
+						
+						//right click on item to hotkey (skillbar)
+						if (spItem->pItemBasic->byClass == ITEM_CLASS_POTION) //usable items
```
twostars: My biggest issue is here. You're assuming that these are the only type of items that can be moved to the skillbar.

Coincidentally, the very first thing I grabbed from Sundries was an Acid Potion -- which does *not* use this type, but *can* still be attached.

Rather than try to reinvent the wheel here with the logic, you should reuse/share the existing logic it uses to identify which items can be dragged onto the bar. Then it will behave consistently for both.

twostars (changes requested):

PR #311, opened 2025-07-10
UITreeSkillDlg - Update to tooltip on skill hover

reignsprime:
## Pull request type

- [ ] Bugfix
- [x] Feature
- [ ] Build related changes

Multiple strings in the UIF are not being used in the tooltip when you hover over a skill.

## What is the new behaviour?
All of the strings are being used and the data is being called from the right location in tbl, even though some of the data in the tbl file is wrong. The wrong data in the tbl file has not been updated, eg. skill descriptions - example below for Absolute Power.

image

To align with the official client.

## Demo
Basic - shows level, not skill points
basic

2nd/ Master - shows skill points
master

Note:  many skill descriptions are wrong or mismatched in the skill_magic_main_us.tbl file. These have not been changed as part of this PR, but explains why the wrong descriptions are shown in the screenshots above.

## Checklist


Client/WarFare/UISkillTreeDlg.h
```diff
@@ -18,10 +18,17 @@ const int SKILL_DEF_SPECIAL1 = 2;
 const int SKILL_DEF_SPECIAL2 = 3;
 const int SKILL_DEF_SPECIAL3 = 4;
 
-// MAX_SKILL_KIND_OF	:						 Base Skill 1개, 전문 스킬 4개..			
-// MAX_SKILL_IN_PAGE	:						 한 페이지 내의 아이콘 갯수..				
-// MAX_SKILL_PAGE_NUM	:						 스킬 한개가 가지는 최대 페이지수..		
-
+// Master Scrolls (items consumed)
+constexpr DWORD ITEM_ID_MASTER_SCROLL_WARRIOR		= 379063000;
```
twostars: Just use uint32_t here for future's sake, rather than unnecessarily keeping it Microsoft-specific. The struct it's being checked against should already be using uint32_t, so it keeps it consistent.

twostars: Additionally, defining special item IDs here is probably a bad idea - I'd put them in GameDef.h, since they'll most likely need access to other things (e.g. CMagicSkillMng) for its own validation.

Client/WarFare/UISkillTreeDlg.h
```diff
@@ -18,10 +18,17 @@ const int SKILL_DEF_SPECIAL1 = 2;
...
+constexpr DWORD ITEM_ID_MASTER_SCROLL_ROGUE		= 379064000;
+constexpr DWORD ITEM_ID_MASTER_SCROLL_MAGE		= 379065000;
+constexpr DWORD ITEM_ID_MASTER_SCROLL_PRIEST		= 379066000;
+
+// Skill IDs requiring dual/double weapons
+constexpr int SKILL_ID_DUAL_WEAPON_WARRIOR		= 1055;
```
twostars: These aren't exactly 'skill IDs' as we know them. This is a confusing name in this respect.
A 'skill ID' is a 6 digit ID referencing a skill in the various magic tables.

I understand it's used by `iNeedSkill` but a distinction needs to be made here for clarity.

Client/WarFare/UISkillTreeDlg.cpp
```diff
@@ -1036,57 +1051,92 @@ void CUISkillTreeDlg::TooltipRenderEnable(__IconItemSkill* spSkill)
...
+		m_pStr_skill_item2->SetVisible(true);
+
+	switch (spSkill->pSkill->dwExhaustItem)
+	{
+		case ITEM_ID_MASTER_SCROLL_WARRIOR:
+		case ITEM_ID_MASTER_SCROLL_ROGUE:
+		case ITEM_ID_MASTER_SCROLL_MAGE:
+		case ITEM_ID_MASTER_SCROLL_PRIEST:
+		{
+			__TABLE_ITEM_BASIC* pItem = CGameBase::s_pTbl_Items_Basic.Find(spSkill->pSkill->dwExhaustItem - 4000);		// Master stones based on master scroll
```
twostars: This logic is very hardcoded here, and assumes the IDs are very specific and relate to one another directly. I would restructure this slightly, such that I'd define the ID for the stones too, update the switch to map them to the ID for the stones (but decouple it from the notion of it being for the stones specifically), and then after the loop, handle fetching and showing the name based on the required item.

That way it all maps directly and doesn't make any assumptions, and isn't tied to just the scrolls/stones.

Client/WarFare/resource.h
```diff
@@ -271,29 +271,34 @@
...
+#define IDS_SKILL_TOOLTIP_NEED_ITEM_ID2 4312	// Required Item : Dagger
+#define IDS_SKILL_TOOLTIP_NEED_ITEM_ID3 4313	// Required Item : Staff
+#define IDS_SKILL_TOOLTIP_NEED_ITEM_ID4 4314	// Required Item : Arrow
+#define IDS_SKILL_TOOLTIP_NEED_ITEM_ID5 4315	// Required Item : Javelin
+#define IDS_SKILL_TOOLTIP_NEED_ITEM_ID6 4316	// Required Item : Spear
+#define IDS_SKILL_TOOLTIP_NEED_ITEM_ID7 4317	// Required Item : Shield
+#define IDS_SKILL_TOOLTIP_NEED_ITEM_ID8 4318	// Required Item : Bow
+#define IDS_SKILL_TOOLTIP_NEED_ITEM_ID9 4319	// Required Item : Longbow
+#define IDS_SKILL_TOOLTIP_MANA_NO	    	4320	// Doesn't use MP						
+#define IDS_SKILL_TOOLTIP_STONE_ITEM		4321	// Item consumed : %s
```
twostars: These are named such for a specific use case: the stones.
But this isn't representative of the resource itself.

Rename this to be more generic.

Client/WarFare/resource.h
```diff
@@ -271,29 +271,34 @@
...
+#define IDS_SKILL_TOOLTIP_STONE_NO			4322	// No item consumed
```
twostars: For reference, this one too:

It's also named such for a specific use case: the stones.
But this isn't representative of the resource itself.

Rename this to be more generic.

Client/WarFare/UISkillTreeDlg.cpp
```diff
@@ -963,71 +963,86 @@ void CUISkillTreeDlg::ButtonTooltipRender(int iIndex)
 	m_pStr_info->Render();
 }
 
+// Render skill tooltip on skill hover
 void CUISkillTreeDlg::TooltipRenderEnable(__IconItemSkill* spSkill)
 {
-	if (spSkill == nullptr
-		|| spSkill->pSkill == nullptr)
+	if (spSkill == nullptr || spSkill->pSkill == nullptr)
```
twostars: Unnecessary reformat here.

twostars (changes requested):

Client/WarFare/UISkillTreeDlg.h
reignsprime: Changes name and comment to reflect they are skills that require a dual or double weapon to use, not the skill ID itself.

Client/WarFare/UISkillTreeDlg.cpp
reignsprime: Thanks for the feedback, I need to change my mindset from just working on 1298 functionality to possible future uses and maintaining the code.

PR #312, opened 2025-07-10
Replace DB access with nanodbc + OpenKO-db-modules

### LINKED ISSUE #252: Replacing the database implementation logic
With #248 we'll be going back to the official MFC recordsets (over ODBC) + manual ODBC code.
Before we start introducing additional tables and stored procedures, we should discuss where we want to take this.

This is the perfect time to update it to use a more modernised library. This would be especially helpful for reducing how much effort is required to implement individual requests (e.g. in Aujard or VersionManager), where you have exceptionally unwieldy code like:
```cpp
int CDBAgent::AccountLogout(const char* accountid, int iLogoutCode)
{
	SQLHSTMT		hstmt = nullptr;
	SQLRETURN		retcode;
	TCHAR			szSQL[1024] = {};
	SQLSMALLINT		sParmRet = 0;
	SQLINTEGER		iUnk = 0;
	SQLINTEGER		cbParmRet = SQL_NTS;

	wsprintf(szSQL, TEXT("{call ACCOUNT_LOGOUT( \'%hs\', %d, ?, ?)}"), accountid, iLogoutCode);

	DBProcessNumber(19);

	CTime t = CTime::GetCurrentTime();
	char strlog[256] = {};
	sprintf(strlog, "[AccountLogout] acname=%s \r\n", accountid);
	m_pMain->WriteLogFile(strlog);
	//m_pMain->m_LogFile.Write(strlog, strlen(strlog));
	TRACE(strlog);

	retcode = SQLAllocHandle(SQL_HANDLE_STMT, m_AccountDB.m_hdbc, &hstmt);
	if (retcode == SQL_SUCCESS)
	{
		retcode = SQLBindParameter(hstmt, 1, SQL_PARAM_OUTPUT, SQL_C_SSHORT, SQL_SMALLINT, 0, 0, &sParmRet, 0, &cbParmRet);
		retcode = SQLBindParameter(hstmt, 2, SQL_PARAM_OUTPUT, SQL_C_SLONG, SQL_INTEGER, 0, 0, &iUnk, 0, &cbParmRet);
		if (retcode == SQL_SUCCESS)
		{
			retcode = SQLExecDirect(hstmt, (SQLTCHAR*) szSQL, SQL_NTS);
			if (retcode == SQL_SUCCESS
				|| retcode == SQL_SUCCESS_WITH_INFO)
			{
				SQLFreeHandle(SQL_HANDLE_STMT, hstmt);
				if (sParmRet == 0)
					return -1;
				else
					return sParmRet; // 1 == success
			}
			else
			{
				if (DisplayErrorMsg(hstmt) == -1)
				{
					m_AccountDB.Close();
					if (!m_AccountDB.IsOpen())
					{
						ReConnectODBC(&m_AccountDB, m_pMain->m_strAccountDSN, m_pMain->m_strAccountUID, m_pMain->m_strAccountPWD);
						return FALSE;
					}
				}

				SQLFreeHandle(SQL_HANDLE_STMT, hstmt);
				return -1;
			}
		}

		SQLFreeHandle(SQL_HANDLE_STMT, hstmt);
	}

	return -1;
}
```

Which you'll note has several flaws, ESPECIALLY considering none of them (except the one I added to VersionManager to replace the ACCOUNT_LOGIN implementation) bind their input.

I don't have any specific options at present, but for it to be viable it would have to be a mature project, reasonably efficient, and modern.
It wouldn't *necessarily* have to use ODBC; it could connect to MSSQL directly, though this would change our configuration.

Suggestions and input is welcome here.

twostars:
As discussed in #252, we're looking to replace our database implementation.

Currently we use MFC for recordsets, with handcrafted structs as our table models (or typically, a subset of the table anyway).
Additionally, particularly for Aujard and VersionManager, we use the native ODBC API for interacting with the database server.

We intend to change this over to use [nanodbc](https://github.com/nanodbc/nanodbc), combined with [models generated from](https://github.com/Open-KO/OpenKO-db-modules) and for the [real database schema](https://github.com/Open-KO/OpenKO-db) via our [kodb-util](https://github.com/Open-KO/kodb-util) and related projects.

What this effectively means is that any schema changes will be directly reflected in any and all of the projects' related structs without any additional user involvement, while they all remain consistently documented throughout.

This PR is a work in progress.

Currently we:
 - [x] Generate full documented table models.
 - [x] Generate server-specific subsets of said models as needed.
 - [x] Fetch and build nanodbc.
 - [x] Fetch and build OpenKO-db-modules (dump of our models, in C++20 module form).
 - [x] Import nanodbc in all related server projects.
 - [x] Import and correctly access the DB model modules in all related server projects.
 - [x] Replace existing recordset code with our own (we have a mockup, which will be integrated).
   - [x] AI server
   - [x] Ebenezer
   - [x] Aujard
   - [x] ~~VersionManager~~ (not applicable)
 - [x] Replace existing stored procedure code with our own (no mockup for this yet).
   - [x] Aujard (mostly -- needs fixed binary bindings)
   - [x] VersionManager
 - [x] Implement connection pooling managed entirely by db::Connectionmanager to replace the manual individual connections in the applications (e.g. Aujard's account 1, game 1, game 2, for the sake of separate thread use).
       Ideally we shouldn't ever need to care about the implementation details - we just want it to run on a database, so get us a connection for that database.
 - [x] Shift out db-library code into a separate submodule to have built via our existing fetch strategy, to avoid it being unnecessarily rebuilt every build.
 - [x] Before final build, remove the WIZ_WAREHOUSE/WAREHOUSE_OPEN fix. This should go in a separate PR. It's just there so that we can verify items are loaded and saving correctly.

Additionally, it should be noted that we move all project intermediate directories to .int/ [under the root directory] for consistency.
This makes them easier to find (and delete for quickly freeing up space), and we can just more easily access it for the sake of C++20 module searches.
This has the effect that existing repos may end up appearing to be considerably larger on next build, so if space is a concern, either re-clone or hunt down and delete all the .int/ folders under each project (leave the one in the root directory though).

Finally, we also just update the server projects to use C++23.
This is needed to make life simpler regarding importing modules from other projects, otherwise we're forced to import them individually at a file level, instead of it being able to find them automatically.

Closes #252


.gitmodules
```diff
@@ -11,3 +11,9 @@
 [submodule "deps/libjpeg-turbo"]
 	path = deps/libjpeg-turbo
 	url = https://github.com/libjpeg-turbo/libjpeg-turbo
+[submodule "deps/nanodbc"]
+	path = deps/nanodbc
+	url = https://github.com/nanodbc/nanodbc
```
kenner2: we should snapshot at a commit hash / version tag so that something being added to the bleeding edge doesn't introduce a break we're not expecting

twostars: It's tied to a commit at the submodule's directory.

kenner2:
## Aujard testing state:
Given the state of the server prior to this PR, some of these may be OK to merge in a failing state.

- [x] AccountLogIn
- [x] CreateNewChar
- [ ] SelectNation
-> works fine for existing accounts, haven't tested new account yet
- [x] SelectCharacter
- [x] Save Userdata (logout, /save, zone change)
-> Working without errors but possible issues given LoadCharInfo errors
- [x] UserLogOut
- [ ] Kick user
- [ ] Observe CURRENTUSER behavior
- [ ] Behavior when Ebenezer crashes (or is closed -> AllSaveRoutine)
- [ ] LoadKnightsAllList (Ranking/War)
- [ ] Battle Event (Kill commander in war)

Clan management:
- [ ] create
- [ ] add member (accept/reject)
- [ ] member leave
- [ ] kick member
- [ ] promote member
- [ ] disband

## observations
_dbAgent.LoadCharInfo(): ..\nanodbc\nanodbc\nanodbc.cpp:4008: 24000: [Microsoft][ODBC SQL Server Driver]Invalid cursor state 

warehouse data seems to be getting bricked after logout. a lot of the following errors logged:
LoadWarehouseData(): item dropped itemId=3452816640 accountId=testing
LoadWarehouseData(): item dropped itemId=3452816845 accountId=testing

AccountLogout proc not binding output params:
AccountLogout(): ret1 not updated by proc for accountId=kenner
this doesn't affect the proc from working but will be problematic for other procs

twostars:
> _dbAgent.LoadCharInfo(): ..\nanodbc\nanodbc\nanodbc.cpp:4008: 24000: [Microsoft][ODBC SQL Server Driver]Invalid cursor state

I inadvertently fixed by not bothering to even fetch empty names (since it requests all 3 no matter what).

> warehouse data seems to be getting bricked after logout. a lot of the following errors logged:
> LoadWarehouseData(): item dropped itemId=3452816640 accountId=testing
> LoadWarehouseData(): item dropped itemId=3452816845 accountId=testing

Needs to be bound as binary, the procs are generated exposing it directly as null-terminated string data. Since nanodbc doesn't let us even provide a length, we need to regenerate these to use `const std::vector<uint8_t>&` for these intended-as-binary fields, so it can be passed through appropriately (and then it has the vector's size to use).

> AccountLogout proc not binding output params:
> AccountLogout(): ret1 not updated by proc for accountId=kenner
> this doesn't affect the proc from working but will be problematic for other procs

Output variables & return values are only fetched once it's finished reading all other result sets.
I enforce this with the flush_on_destruct() call in the stored procedures, so the simplest thing is to just check them outside of scope. Since the database stuff is scoped by try..catch blocks, with everything else it then makes sense to just check this after the try..catch block, rather than within it. By that point it'll have updated.

I fixed this.

twostars:
@sourcery-ai: review

twostars:
@sourcery-ai review

PR #315, opened 2025-07-12
code cleanup for ui transaction

koraykarakus:
## Pull request type

## What is the new behaviour?

* NULL macro replaced with nullptr
* N3_VERIFY_UI_COMPONENT macro is used on required places.
* Comments translated from Korean to English ( by AI )
* predefination of i,j,k moved inside loops as local scope
* FAIL_RETURN inline macro replaced with lambda

To obtain more readability.

## Checklist

## Summary by Sourcery

Modernize UITransactionDlg code style and enhance readability

Enhancements:
- Replace NULL with nullptr throughout the UI transaction code
- Declare loop indices locally inside their loops for tighter scope
- Replace existing FAIL_RETURN macro with a lambda-based early-return mechanism
- Use N3_VERIFY_UI_COMPONENT macro for consistent UI component validation
- Translate Korean comments into English for improved clarity

### DIFF
--- a/Client/WarFare/UITransactionDlg.cpp
+++ b/Client/WarFare/UITransactionDlg.cpp
@@ -35,20 +35,21 @@ static char THIS_FILE[]=__FILE__;
 
 CUITransactionDlg::CUITransactionDlg()
 {
-	int i, j;
 	m_iCurPage = 0;
-	for( j = 0; j < MAX_ITEM_TRADE_PAGE; j++ )
-		for( i = 0; i < MAX_ITEM_TRADE; i++ )		
-			m_pMyTrade[j][i] = NULL;
-	for( i = 0; i < MAX_ITEM_INVENTORY; i++ )	m_pMyTradeInv[i] = NULL;
+	
+	for(int i = 0; i < MAX_ITEM_TRADE_PAGE; i++ )
+		for(int j = 0; j < MAX_ITEM_TRADE; j++ )		
+			m_pMyTrade[i][j] = nullptr;
 
-	m_pUITooltipDlg = NULL;
-	m_pStrMyGold    = NULL;
+	for(int i = 0; i < MAX_ITEM_INVENTORY; i++ )	
+		m_pMyTradeInv[i] = nullptr;
 
-	m_pUIInn		= NULL;
-	m_pUIBlackSmith	= NULL;
-	m_pUIStore		= NULL;
-	m_pText_Weight = nullptr;
+	m_pUITooltipDlg = nullptr;
+	m_pStrMyGold    = nullptr;
+	m_pUIInn		= nullptr;
+	m_pUIBlackSmith	= nullptr;
+	m_pUIStore		= nullptr;
+	m_pText_Weight	= nullptr;
 
 	this->SetVisible(false);
 }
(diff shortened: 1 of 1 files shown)


Client/WarFare/UITransactionDlg.cpp
```diff
@@ -772,39 +769,41 @@ void CUITransactionDlg::ReceiveItemDropByTradeSuccess()
...
 bool CUITransactionDlg::ReceiveIconDrop(__IconItemSkill* spItem, POINT ptCur)
 {
-// Temp Define 
-#define FAIL_RETURN {	\
-		CN3UIWndBase::AllHighLightIconFree();	\
-		SetState(UI_STATE_COMMON_NONE);	\
-		return false;	\
-	}
+	// Temp Define 
+	auto FAIL_RETURN = [&]() -> bool
```
twostars: These are not equivalent.
Returning here returns from the lambda. Execution carries on immediately after.

Previous behaviour caused ReceiveIconDrop() to return false in this case.

twostars (changes requested):
I strongly disagree with relying on AI to translate these comments.

This is the 3rd PR I've had to go through and manually review and retranslate & localise to ensure the original intent from the comments is preserved.

A misleading or outright wrong comment is *far* worse than one most people can't read, because the original is at least still intact with its original intent for proper, sensible translation.

Unless you can properly review the comments and verify their accuracy, I recommend not AI translating them at all. It just does more harm than good. I'm just having to spend time going over it myself every time to catch its mistakes.

Client/WarFare/UITransactionDlg.cpp
koraykarakus: true I forgot that, need to check if it is false and exterminate if it is false, but it adds more complication.

I tried to change that because if statements with paranthesis {} can cause macro to fail. maybe macro should be updated and should be placed inside do scope.

twostars: Just replace the macro with its implementation logic. It's the simplest option.

It's minor code duplication, but it's still cleaner than the mess it is with the macro. From there the code can be shuffled and refactored to reduce the introduced duplication.

koraykarakus: I understand the importance of preserving comments, especially to keep track of the original developers' intent. However, considering that the game isn't even 30% functional at this point, I think our priority should be improving and stabilizing the code.

Also, many comments are in Korean, and in several cases, they don’t provide meaningful insight—some just state what's already obvious from the variable or function names.

For example, when I’m trying to improve or refactor a part of the code, a Korean comment often ends up adding confusion rather than clarity. Translating these takes a lot of time and usually doesn’t add much value. In my opinion, in the current state of the project, clean and functional code should come before preserving low-quality or redundant comments.

twostars: You're just making the argument for not bothering with touching the comments in the first place.
You want this to be readable, but when your comments are inaccurate/misleading, in some cases being just COMPLETELY wrong and not matching what the original comment said at all, this just has the opposite effect.

All it does is it causes unnecessary confusion, just because you wanted to let AI translate the comments without bothering to proofread them.

koraykarakus: First of all, I haven't used any external API for comment translation. So claiming I didn't do any proofreading isn't accurate — even unintentionally, I did perform a kind of manual review while translating.

Second, even if some translated comments aren’t perfect, having them in English opens the door for future contributors to quickly understand and improve them — without needing any external tools or knowledge of Korean.
In contrast, keeping comments in Korean — in a project where not a single contributor speaks Korean — just creates unnecessary friction. Anyone trying to understand that part of the code has to waste time translating it themselves, adding zero value to the actual development process.

The purpose of comments is to help people understand the code. If the comments themselves require translation before they can be useful, then they’re not serving their purpose — they're just clutter.

So even if a translated comment is imperfect, it’s still far more helpful and future-proof than a Korean one that no one here can even read.

twostars:
Okay, I'm going to close this one.

What I want to do here is setup `clang-format` to consistently handle formatting so we can just run a pass on these files instead of doing it manually, as is done here, which only targets a couple of things.

The 3 main concerns in this commit are:
1. The comment translations.
While I generally don't care too much about the comments, I also would prefer they weren't translated to be misleading; reading a misleading comment that makes you misunderstand the code is worse than simply not being able to understand the original comment (for those that don't understand Korean) which is far more likely to be correct (though not always).
In this case, it's better to make your own judgement based on what the code actually does than defer to the misleading comment.
If they're to be changed, they should be done so in a thoughtful manner, with care to ensure accuracy.

Obviously not all comments are even worth keeping; a lot of them are obvious. But we shouldn't assume that's the case for them all.

2. Flipping of the loop variables; while I can't find fault with them, they do make me very hesitant in case any of it was missed, causing obscure bugs.

3. The FAIL_RETURN stuff which is actually entirely useless; if we follow the logic back, we find that this behaviour is automatic anyway, so it could just return false and it doesn't even need to handle this specific logic.

But that aside, we really don't need a lambda for this anyway; it's simple enough and used so infrequently it could've just been called manually.
That, combined with obscurely returning whatever the lambda returns (which is false) makes things less readable in my eyes, than the macro (somehow). In this case, the lambda should have simply returned void, and the caller should've returned false explicitly. That would've been clearer and readable.

Returning a "dynamic" result implies it might not even return false. You could assume it from it being called "fail_return", but you don't know for sure, specifically because it's left to fail_return to return it.

Regardless, I'll just close this PR for now. An eventual cleanup pass with clang-format will suffice for this.

PR #330, opened 2025-07-25
Implement spdlog as universal logger in server applications

kenner2:
Implement spdlog as universal logger in server applications

## Pull request type
- [x] Feature

Server applications uses different logger functions and formats, all of which are likely blocking the main thread when logging to file needs to occur.

## What is the new behaviour?
Server applications and deps will use spdlog for logging.  spdlog will be configured for both stdout for console-based logging as well as make use of daily log files.  .ini files will be updated to support logger configuration.

## Checklist


UTengine:
Sxy

Server/Ebenezer/EbenezerDlg.cpp
```diff
@@ -389,163 +376,181 @@ BOOL CEbenezerDlg::OnInitDialog()
...
+		spdlog::error("EbenezerDlg::OnInitDialog: failed to connect to the AIServer");
 #ifndef _DEBUG
 		AfxPostQuitMessage(0);
 #endif
 	}
 
-	LogFileWrite(_T("success\r\n"));
-	UserAcceptThread();
+	// this is called after all data is synced with AIServer - I don't think it should be called here.
+	// UserAcceptThread();
```
kenner2: @twostars just to call this out.  Everything worked OK with it commented out.

twostars: I think it's like this for debug builds. It's meant to be called when AI finishes connecting, to allow users to connect.
With it existing like this it means players can connect before AI loads up, which is nicer for debug (especially when you don't want to go through the trouble of loading up AI to test something).

deps/spdlog-msvc/spdlog.vcxproj
```diff
@@ -0,0 +1,232 @@
+﻿<?xml version="1.0" encoding="utf-8"?>
+<Project DefaultTargets="Build" ToolsVersion="17.0" xmlns="http://schemas.microsoft.com/developer/msbuild/2003">
+  <PropertyGroup>
+    <PreferredToolArchitecture>Win32</PreferredToolArchitecture>
+  </PropertyGroup>
+  <PropertyGroup Condition="'$(Configuration)|$(Platform)'=='Debug|x64'">
```
twostars: This block can go

deps/spdlog-msvc/spdlog.vcxproj
```diff
@@ -0,0 +1,232 @@
...
+    <Import Project="$(UserRootDir)\Microsoft.Cpp.$(Platform).user.props" Condition="exists('$(UserRootDir)\Microsoft.Cpp.$(Platform).user.props')" Label="LocalAppDataPlatform" />
+  </ImportGroup>
+  <PropertyGroup Label="UserMacros" />
+  <Import Project="$(PropsDir)spdlog.props" />
+  <ItemDefinitionGroup>
+    <ClCompile>
+      <PrecompiledHeader>NotUsing</PrecompiledHeader>
+      <PreprocessorDefinitions>%(PreprocessorDefinitions);</PreprocessorDefinitions>
+      <LanguageStandard>stdcpp20</LanguageStandard>
+      <AdditionalIncludeDirectories>$(DependencyDir)\spdlog\include;%(AdditionalIncludeDirectories)</AdditionalIncludeDirectories>
```
twostars: 2 things:
1. $(DependencyDir) (like all paths here) ends in a \, so you drop the \ here.
2. It's better to just have the include directory here in spdlog.props (which it probably already is) since it'll be exposed to others.

deps/spdlog-msvc/spdlog.vcxproj
```diff
@@ -0,0 +1,232 @@
...
+  <ImportGroup Label="PropertySheets">
+    <Import Project="$(UserRootDir)\Microsoft.Cpp.$(Platform).user.props" Condition="exists('$(UserRootDir)\Microsoft.Cpp.$(Platform).user.props')" Label="LocalAppDataPlatform" />
+  </ImportGroup>
+  <PropertyGroup Label="UserMacros" />
+  <Import Project="$(PropsDir)spdlog.props" />
+  <ItemDefinitionGroup>
+    <ClCompile>
+      <PrecompiledHeader>NotUsing</PrecompiledHeader>
+      <PreprocessorDefinitions>%(PreprocessorDefinitions);</PreprocessorDefinitions>
+      <LanguageStandard>stdcpp20</LanguageStandard>
```
twostars: Unless it doesn't explicitly support it, I would just bump this up to C++23. I know fmtlib at least enables certain behaviour that's only possible with C++23, so spdlog should probably also do it by proxy.

Unless it just does not support it for whatever reason, of course.

Server/AIServer/GameSocket.cpp
```diff
@@ -272,11 +272,7 @@ void CGameSocket::RecvUserInfo(char* pBuf)
 	if (sLength > MAX_ID_SIZE
 		|| sLength <= 0)
 	{
-		CTime cur = CTime::GetCurrentTime();
-		CString logstr;
-		logstr.Format(_T("RecvUserInfo() Fail : %d월 %d일 %d시 %d분 - uid=%d, name=%hs\r\n"), cur.GetMonth(), cur.GetDay(), cur.GetHour(), cur.GetMinute(), uid, strName);
-		LogFileWrite(logstr);
-		TRACE(_T("###  RecvUserInfo() Fail ---> uid = %d, name=%hs  ### \n"), uid, strName);
+		spdlog::error("CGameSocket.RecvUserInfo: charId len={} overflow for userId={}", sLength, uid);
```
twostars: This seems really inconsistent - in some places you use `<full class name>.<method name>`, in others you use `<method name>`, in others still you use `<class name without the "C" prefix>::<method name>`.

Server/AIServer/GameSocket.cpp
```diff
@@ -1104,11 +1108,13 @@ void CGameSocket::RecvHealMagic(char* pBuf)
...
 		{
 			pUser->m_bLive = USER_LIVE;
-			TRACE(_T("##### CGameSocket-RecvHealMagic Fail : User가 Heal된 경우.. [id=%d, %hs, bLive=%d, hp=%d] ######\n"), pUser->m_iUserId, pUser->m_strUserID, pUser->m_bLive, pUser->m_sHP);
+			spdlog::debug("RecvHealMagic: user healed [userId={} charId={} isAlive={} hp={}]",
+				pUser->m_iUserId, pUser->m_strUserID, pUser->m_bLive, pUser->m_sHP);
 		}
 		else
 		{
-			TRACE(_T("##### CGameSocket-RecvHealMagic Fail : UserDead  [id=%d, %hs, bLive=%d, hp=%d] ######\n"), pUser->m_iUserId, pUser->m_strUserID, pUser->m_bLive, pUser->m_sHP);
+			spdlog::error("RecvHealMagic:  user is dead [userId={} charId={} isAlive={} hp={}]",
```
twostars: It seems like it, but I think this should be a warning, not an error.
The reason being - order of operation. It can receive these requests from Ebenezer (where they were alive at the time) after the user has died to an NPC (in this server). This kind've thing just happens.

Server/AIServer/MagicProcess.cpp
```diff
@@ -5,6 +5,9 @@
 #include "stdafx.h"
 #include "server.h"
 #include "MagicProcess.h"
+
+#include <spdlog/spdlog.h>
```
twostars: I see what you're doing here by trying to order it alphabetically, but typically these are separated so it's like, local includes first, then the external dependencies

Server/AIServer/Npc.cpp
```diff
@@ -639,8 +646,7 @@ void CNpc::NpcTracing(CIOCPort* pIOCP)
 
 void CNpc::NpcAttacking(CIOCPort* pIOCP)
 {
-	if (cur_test)
-		NpcTrace(_T("NpcAttacking()"));
+	NpcTrace("NpcAttacking()");
```
twostars, line 650: I don't really know what to think about these.

You have to remember, the server is very busy, spamming these things constantly across numerous threads.

Before it was basically optimised out entirely because that was never set, so we didn't get the hit from it.
Now it triggers all the time for every NPC.

I think it needs a compile time flag back in some capacity, even if it's under NpcTrace() directly, otherwise this is just going to be a mess of logs.

Server/AIServer/RoomEvent.cpp
```diff
@@ -5,6 +5,9 @@
 #include "stdafx.h"
 #include "server.h"
 #include "RoomEvent.h"
+
+#include <spdlog/spdlog.h>
```
twostars: Again, this one's ordered alphabetically but typically external includes go below the local includes.

Server/AIServer/ServerDlg.cpp
```diff
@@ -4,15 +4,20 @@
 #include "stdafx.h"
 #include "Server.h"
 #include "ServerDlg.h"
+
+#include <codecvt>
```
twostars: External dependency - move below local

Server/AIServer/ServerDlg.cpp
```diff
@@ -366,26 +366,31 @@ BOOL CServerDlg::OnInitDialog()
 	//----------------------------------------------------------------------
 	//	Start Accepting...
 	//----------------------------------------------------------------------
-	if (m_byZone == KARUS_ZONE
+	if ((m_byZone == KARUS_ZONE
```
twostars: Better to make this a separate method at this point.
This repeat logic is horrible.

The formatting is wonky here too, with `&&!m_Iocport.Listen`

Server/AIServer/ServerDlg.cpp
```diff
@@ -479,7 +485,8 @@ BOOL CServerDlg::GetMagicTableData()
 		ReportTableLoadError(loader.GetError(), __func__);
 		return FALSE;
 	}
-
+	
+	spdlog::info("MAGIC cached");
```
twostars: I would change these to 'loaded' (every one of these is a LoadXyzTable method).

Server/AIServer/ServerDlg.cpp
```diff
@@ -782,8 +800,9 @@ BOOL CServerDlg::LoadNpcPosTable(std::vector<model::NpcPos*>& rows)
 
 					if (row->RespawnTime < 15)
 					{
-						TRACE(_T("##### ServerDlg:CreateNpcThread - RegenTime Error :  nid=%d, name=%hs, regentime=%d #####\n"), pNpc->m_sNid + NPC_BAND, pNpc->m_strName, row->RespawnTime);
-						row->RespawnTime = 30;
+						spdlog::warn("LoadNpcPosTable: RegTime below minimum value of 15s [npcId={}, serial={}, npcName={}, RegTime={}]",
+							pNpc->m_sSid, pNpc->m_sNid + NPC_BAND, pNpc->m_strName, row->RespawnTime);
+						row->RespawnTime = 15;
```
twostars: This is a very subtle change from official behaviour. This continues to reset this 30 in the official 1.298 binary.

Server/AIServer/ServerDlg.cpp
```diff
@@ -796,18 +815,18 @@ BOOL CServerDlg::LoadNpcPosTable(std::vector<model::NpcPos*>& rows)
...
 								pNpc->m_sSid,
-								pNpc->m_strName.c_str(),
+								pNpc->m_strName,
 								pNpc->m_byMoveType,
 								pNpc->m_sMaxPathCount);
-							TRACE(error);
-							AfxMessageBox(error);
 
+							spdlog::error(error);
+							std::wstring_convert<std::codecvt_utf8<wchar_t>> converter;
```
twostars: We have existing wrapper code for "this", though it's Windows-centric right now, but it can be swapped out later as needed.
See: shared\StringConversion.h

*Technically* right now it should be CpToWide(error, 949) for Korean, but LocalToWide() is sufficient. The names are probably more likely to be encoded using their local encoding right now instead of utf8, and I suspect any Korean is just mangled in the DB.

Server/AIServer/ServerDlg.cpp
```diff
@@ -824,18 +843,18 @@ BOOL CServerDlg::LoadNpcPosTable(std::vector<model::NpcPos*>& rows)
...
 								row->PathPointCount,
 								pNpc->m_sNid + NPC_BAND,
 								pNpc->m_sSid,
 								pNpc->m_strName.c_str(),
 								pNpc->m_byMoveType,
 								pNpc->m_sMaxPathCount);
-							TRACE(error);
-							AfxMessageBox(error);
+							spdlog::error(error);
+							std::wstring_convert<std::codecvt_utf8<wchar_t>> converter;
```
twostars: We have existing wrapper code for "this", though it's Windows-centric right now, but it can be swapped out later as needed.
See: shared\StringConversion.h

*Technically* right now it should be CpToWide(error, 949) for Korean, but LocalToWide() is sufficient. The names are probably more likely to be encoded using their local encoding right now instead of utf8, and I suspect any Korean is just mangled in the DB.

Server/AIServer/ServerDlg.cpp
```diff
@@ -883,15 +902,16 @@ BOOL CServerDlg::LoadNpcPosTable(std::vector<model::NpcPos*>& rows)
...
+						AfxMessageBox(_T("NPC invalid zone index error (see log)"));
 						return FALSE;
 					}
 
 					//pNpc->Init();
 					//m_arNpc.Add(pNpc);
 					if (!m_arNpc.PutData(pNpc->m_sNid, pNpc))
 					{
-						TRACE(_T("Npc PutData Fail - %d\n"), pNpc->m_sNid);
+						spdlog::warn("Npc PutData Fail - {}", pNpc->m_sNid);
```
twostars: This should be an error; it's *absolutely* not intended for an NPC to fail to load.
I'm pretty sure you used error for this in other cases.

Server/AIServer/ServerDlg.cpp
```diff
@@ -1153,8 +1166,11 @@ BOOL CServerDlg::MapFileLoad()
 			CFile file;
 			if (!file.Open(szFullPath, CFile::modeRead))
 			{
-				errormsg.Format(_T("파일 Open 실패 - %s\n"), szFullPath.GetString());
-				AfxMessageBox(errormsg);
+				std::string error = std::format("MapFileLoad: Failed to open file: {}", mapPath.generic_string());
```
twostars: Okay, because this one's a path it gets messy. But here's what you can do.
Make the base error a wide string. Fetch the path as a native_string() (which is std::wstring) and convert the entire error with WideToUtf8() from shared/StringConversion.h

Then pass the utf-8 error to spdlog, and pass the wide error to AfxMessageBox.

Server/AIServer/ServerDlg.cpp
```diff
@@ -1164,8 +1180,11 @@ BOOL CServerDlg::MapFileLoad()
 
 			if (!pMap->LoadMap(file.m_hFile))
 			{
-				errormsg.Format(_T("Map Load 실패 - %s\n"), szFullPath.GetString());
-				AfxMessageBox(errormsg);
+				std::string error = std::format("MapFileLoad: Failed to load map file: {}", mapPath.generic_string());
```
twostars: Okay, because this one's a path it gets messy. But here's what you can do.
Make the base error a wide string. Fetch the path as a native_string() (which is std::wstring) and convert the entire error with WideToUtf8() from shared/StringConversion.h

Then pass the utf-8 error to spdlog, and pass the wide error to AfxMessageBox.

Server/AIServer/ServerDlg.cpp
```diff
@@ -1177,8 +1196,11 @@ BOOL CServerDlg::MapFileLoad()
 			{
 				if (!pMap->LoadRoomEvent(row.RoomEvent))
 				{
-					errormsg.Format(_T("Map Room Event Load 실패 - %s\n"), szFullPath.GetString());
-					AfxMessageBox(errormsg);
+					std::string error = std::format("MapFileLoad: LoadRoomEvent failed: {}", mapPath.generic_string());
```
twostars: Okay, because this one's a path it gets messy. But here's what you can do.
Make the base error a wide string. Fetch the path as a native_string() (which is std::wstring) and convert the entire error with WideToUtf8() from shared/StringConversion.h

Then pass the utf-8 error to spdlog, and pass the wide error to AfxMessageBox.

Server/AIServer/ServerDlg.cpp
```diff
@@ -1554,9 +1573,8 @@ void CServerDlg::GameServerAcceptThread()
 
 void CServerDlg::SyncTest()
 {
-	FILE* stream = fopen("c:\\aiserver.txt", "w");
-
-	fprintf(stream, "*****   Check ... List  *****\n");
+
```
twostars: Random newline

Server/Aujard/AujardDlg.cpp
```diff
@@ -465,12 +437,15 @@ void CAujardDlg::SelectCharacter(char* buffer)
...
-		count++;
+		retryCount++;
 	}
-	while (count < 50);
+	while (retryCount < maxRetry);
 
-	if (count >= 50)
+	if (retryCount >= maxRetry)
+	{
 		OutputList.AddString(_T("Packet Drop: WIZ_SEL_CHAR"));
```
twostars: I'm starting to think we should just have a wrapper method for these.
Like, in this case (but for all the servers), something in their main UI class (CAujardDlg, CEbenezerDlg, CServerDlg, CVersionManagerDlg) to write the given string to its listbox (in this case OutputList) with a corresponding log type (presumably you can provide a log type to spdlog, rather than use spdlog::error() directly etc).

At this point it would probably make more sense to flip it so that it's passing a multibyte (we'll consider it utf8) string, then convert it with Utf8ToWide() to output to the list, while forwarding it directly to the logger as-is with the given type.
That would save a lot of the existing conversion cruft.

That said, for filenames and such it might simply make more sense to require wide-strings for it, and then convert it back to utf-8 (WideToUtf8()) for the logger call. There's probably enough of those cases to warrant doing it that way.

Server/Aujard/AujardDlg.cpp
```diff
@@ -929,39 +923,39 @@ void CAujardDlg::AllSaveRoutine()
...
-			TRACE(_T("GameServer Dead!! - %hs Skipped\n"), pUser->m_id);
-#endif
+			spdlog::debug("userId skipped for invalid data: {}", userId);
 			continue;
 		}
 
 		if (HandleUserLogout(userId, UPDATE_ALL_SAVE, true))
 		{
-			TRACE(_T("GameServer Dead!! - %hs Saved\n"), pUser->m_id);
+			spdlog::debug("Character saved: {}", pUser->m_id);
```
twostars: Probably still needs context, just so it's clear why they were saved. There is the initial line above, but it could be saved concurrently with a regular save so there's no way of telling which triggered it.

Server/Aujard/AujardDlg.cpp
```diff
@@ -929,39 +923,39 @@ void CAujardDlg::AllSaveRoutine()
...
 		}
 		else
 		{
 			allUsersSaved = false;
-			TRACE(_T("GameServer Dead!! - %hs Not Saved\n"), pUser->m_id);
+			spdlog::error("Failed to save character: {}", pUser->m_id);
```
twostars: Probably still needs context, just so it's clear why they were attempted to be saved. There is the initial line above, but it could be saved concurrently with a regular save so there's no way of telling which triggered it.

Server/Ebenezer/AISocket.cpp
```diff
@@ -155,20 +154,22 @@ void CAISocket::LoginProcess(char* pBuf)
...
-	CString logstr;
-
+	BYTE zone = GetByte(pBuf, index);
+	// 0: first connect
+	// 1: reconnect
+	BYTE byReConnect = GetByte(pBuf, index);
+	
 	// zone 틀리면 에러 
-	if (ver == -1)
+	if (zone == -1)
```
twostars: I just want to make this clear that while you didn't do this, this doesn't actually work. It's also bugged in the official 1.298 code. They fixed it in later builds, but this comparison doesn't work.
It needs to be 0xff or 255, or for zone itself to be signed (e.g. int8_t).

Server/Ebenezer/AISocket.cpp
```diff
@@ -1055,22 +1063,23 @@ void CAISocket::RecvUserHP(char* pBuf)
...
+	short sLoyalty = GetShort(pBuf, index);
 
-	nid = GetShort(pBuf, index);
-	sExp = GetShort(pBuf, index);
-	sLoyalty = GetShort(pBuf, index);
-
-	CUser* pUser = (CUser*) m_pMain->m_Iocport.m_SockArray[nid];
-	if (pUser == nullptr)
+	CUser* pUser = (CUser*) m_pMain->m_Iocport.m_SockArray[userId];
+	if (pUser == nullptr || pUser->m_pUserData == nullptr)
```
twostars: You don't need to check if m_pUserData is nullptr. It will never be nullptr.
It's initialized on startup and is coupled to the instance.

Server/Ebenezer/EbenezerDlg.cpp
twostars: Essentially I think this should be restored in _DEBUG builds only.

Server/Ebenezer/EbenezerDlg.cpp
```diff
@@ -2341,13 +2319,17 @@ BOOL CEbenezerDlg::LoadNoticeData()
 #if !defined(_DEBUG)
 		AfxMessageBox(_T("cannot open Notice.txt!!"));
 #endif
+		spdlog::error("EbenezerDlg::LoadNoticeData: failed to open Notice.txt");
```
twostars: This actually isn't an error. It intends to run fine without Notice.txt. If you check the caller, it doesn't exit in this case.
This should just be a warning.

Server/Ebenezer/IOCPSocket2.cpp
```diff
@@ -191,7 +192,7 @@ int CIOCPSocket2::Send(char* pBuf, long length, int dwFlag)
 		int last_err = WSAGetLastError();
 		if (last_err == WSA_IO_PENDING)
 		{
-			TRACE(_T("SEND : IO_PENDING[SID=%d]\n"), m_Sid);
+			spdlog::error("CIOCPSocket2::Send: socketId={} IO_PENDING", m_Sid);
```
twostars: This is not an error. This is normal behaviour. It just needs to process.

Server/Ebenezer/IOCPSocket2.cpp
```diff
@@ -200,7 +201,7 @@ int CIOCPSocket2::Send(char* pBuf, long length, int dwFlag)
 		}
 		else if (last_err == WSAEWOULDBLOCK)
 		{
-			TRACE(_T("SEND : WOULDBLOCK[SID=%d]\n"), m_Sid);
+			spdlog::error("CIOCPSocket2::Send: socketId={} WOULDBLOCK", m_Sid);
```
twostars: This is also not an error. This is normal behaviour. It just needs to process.

Server/Ebenezer/IOCPSocket2.cpp
```diff
@@ -273,7 +272,7 @@ int CIOCPSocket2::Receive()
 		}
 		else if (last_err == WSAEWOULDBLOCK)
 		{
-			TRACE(_T("RECV : WOULDBLOCK[SID=%d]\n"), m_Sid);
+			spdlog::error("CIOCPSocket2::Receive: socketId={} WOULDBLOCK", m_Sid);
```
twostars: This is not an error. This is normal behaviour. It just needs to process.

Server/Ebenezer/IOCPSocket2.cpp
```diff
@@ -340,17 +340,14 @@ BOOL CIOCPSocket2::PullOutCore(char*& data, int& length)
...
 
 	if (tmp_buffer[0] != PACKET_START1
 		&& tmp_buffer[1] != PACKET_START2)
 	{
-		TRACE(
-			_T("%d: PullOutCore() - failed to detect header (%02X, %02X)\n"),
-			m_Sid,
-			tmp_buffer[0],
-			tmp_buffer[1]);
+		spdlog::error("{}: PullOutCore() - failed to detect header ({:X}, {:X})",
```
twostars: This should probably explicitly be {:2X}

Server/Ebenezer/IOCPSocket2.cpp
```diff
@@ -370,22 +367,17 @@ BOOL CIOCPSocket2::PullOutCore(char*& data, int& length)
...
 	}
 
 	if (length > buffer_len)
 	{
-		TRACE(
-			_T("%d: PullOutCore() - reported length (%d) is not in buffer (%d) - waiting for now\n"),
-			m_Sid,
-			length,
-			buffer_len);
+		spdlog::error("{}: PullOutCore() - reported length ({}) is not in buffer ({}) - waiting for now",
```
twostars: Not an error. The packet hasn't fully been sent yet, it should receive it in a subsequent notification.

Server/Ebenezer/IOCPSocket2.cpp
```diff
@@ -397,20 +389,16 @@ BOOL CIOCPSocket2::PullOutCore(char*& data, int& length)
 	// We expect a 2 byte tail after the end position.
 	if ((ePos + 2) > buffer_len)
 	{
-		TRACE(
-			_T("%d: PullOutCore() - tail not in buffer - waiting for now\n"),
+		spdlog::error("{}: PullOutCore() - tail not in buffer - waiting for now",
```
twostars: Not an error. The packet hasn't fully been sent yet, it should receive it in a subsequent notification.

Server/Ebenezer/IOCPSocket2.cpp
```diff
@@ -397,20 +389,16 @@ BOOL CIOCPSocket2::PullOutCore(char*& data, int& length)
...
 
 	if (tmp_buffer[ePos] != PACKET_END1
 		|| tmp_buffer[ePos + 1] != PACKET_END2)
 	{
-		TRACE(
-			_T("%d: PullOutCore() - failed to detect tail (%02X, %02X)\n"),
-			m_Sid,
-			tmp_buffer[ePos],
-			tmp_buffer[ePos + 1]);
+		spdlog::error("{}: PullOutCore() - failed to detect tail ({:X}, {:X})",
```
twostars: This should probably be {:2X}

Server/Ebenezer/IOCPort.cpp
```diff
@@ -568,7 +567,7 @@ BOOL CIOCPort::Associate(CIOCPSocket2* pIocpSock, HANDLE hPort)
 {
 	if (hPort == nullptr)
 	{
-		TRACE(_T("ERROR : No Completion Port\n"));
+		spdlog::error("CIOCPort::Associate: received null port");
```
twostars: Should be clear that we're talking about a completion port.

Server/Ebenezer/MagicProcess.cpp
```diff
@@ -5,12 +5,15 @@
 #include "stdafx.h"
 #include "ebenezer.h"
 #include "MagicProcess.h"
+
```
twostars: Newline?

Server/Ebenezer/MagicProcess.cpp
```diff
@@ -1470,14 +1475,13 @@ void CMagicProcess::ExecuteType3(int magicid, int sid, int tid, int data1, int d
 					// Killed by another player.
 					if (!bFlag)
 					{
-					   // 눈싸움전쟁존에서 눈싸움중이라면 공격은 눈을 던지는 것만 가능하도록,,,
+						// make it so that the only possible attack is throwing a snowball
```
twostars: Weird wording; something more like "Players can only attack with snowballs during snow wars" or something.

Server/Ebenezer/UdpSocket.cpp
```diff
@@ -466,32 +474,36 @@ void CUdpSocket::RecvCreateKnights(char* pBuf)
...
+	GetString(charId, pBuf, idlen, index);
 
-	pKnights = m_pMain->m_KnightsArray.GetData(knightsindex);
+	pKnights = m_pMain->m_KnightsArray.GetData(knightsId);
 
 	if (command == KNIGHTS_JOIN)
 	{
-		sprintf(finalstr, "#### %s님이 가입하셨습니다. ####", charid);
+		// TODO: Check this against 1299 ebenezer
+		sprintf(finalstr, "#### %s has joined. ####", charId);
```
twostars, line 489: These strings should be in SERVER_RESOURCE, for reference.

Server/Ebenezer/UdpSocket.cpp
```diff
@@ -631,28 +643,29 @@ void CUdpSocket::RecvModifyFame(char* pBuf, BYTE command)
...
+			knightsId);
 		return;
 	}
 
 	flag = pKnights->m_byFlag;
 
 	// 클랜이나 기사단이 파괴된 메시지를 보내고 유저 데이타를 초기화
 	if (flag == CLAN_TYPE)
-		sprintf(finalstr, "#### %s 클랜이 해체되었습니다 ####", pKnights->m_strName);
+		sprintf(finalstr, "#### %s clan has been disbanded. ####", pKnights->m_strName);
```
twostars: These strings should be in SERVER_RESOURCE, for reference.

Server/Ebenezer/UdpSocket.cpp
```diff
@@ -631,28 +643,29 @@ void CUdpSocket::RecvModifyFame(char* pBuf, BYTE command)
...
 	else if (flag == KNIGHTS_TYPE)
-		sprintf(finalstr, "#### %s 기사단이 해체되었습니다 ####", pKnights->m_strName);
+		sprintf(finalstr, "#### %s knights have been disbanded. ####", pKnights->m_strName);
```
twostars: These strings should be in SERVER_RESOURCE, for reference.

Server/Ebenezer/User.cpp
```diff
@@ -1314,7 +1313,9 @@ void CUser::MoveProcess(char* pBuf)
 		|| m_pUserData->m_sHp == 0)
 	{
 		if (speed != 0)
-			TRACE(_T("### MoveProcess Fail : name=%hs(%d), m_bResHpType=%d, hp=%d, speed=%d, x=%d, z=%d ###\n"), m_pUserData->m_id, m_Sid, m_bResHpType, m_pUserData->m_sHp, speed, (int) m_pUserData->m_curx, (int) m_pUserData->m_curz);
+			spdlog::error("User::MoveProcess: dead user is moving [charId={} socketId={} resHpType={} hp={} speed={} x={} z={}",
```
twostars: This really needs to be wrapped in {}, since there's multiple lines here.

Interestingly this is an error but it's also a typical hack situation. With that in mind I'm not sure if an error log is the right place for this.

Server/VersionManager/IOCPort.cpp
```diff
@@ -61,8 +62,7 @@ DWORD WINAPI AcceptThread(LPVOID lp)
 				if (pSocket == nullptr)
 				{
 					TRACE("Socket Array has Broken...\n");
-					_stprintf(logstr, _T("Socket Array has Broken...[sid:%d]\r\n"), sid);
-					LogFileWrite(logstr);
+					spdlog::info("Socket Array has Broken...[sid:{}]", sid);
```
twostars: This is an error. This is a really bad critical error.
It shouldn't happen in practice, it was only because they didn't guard properly, but still; this is absolutely just an error.

Server/shared/logger.cpp
```diff
@@ -0,0 +1,86 @@
...
+#include <spdlog/sinks/daily_file_sink.h>
+#include <spdlog/common.h>
+#include <spdlog/sinks/stdout_color_sinks.h>
+#include <spdlog/async.h>
+#include <spdlog/async_logger.h>
+
+/// \brief Sets up spdlog from an ini file using standardized server settings
+/// \param ini server application's ini file (already loaded)
+/// \param name application name (VersionManager, Aujard, AIServer, Ebenezer)
+void SetupLogger(CIni& ini, const std::string& name)
```
twostars: I feel like this shouldn't just sit in the global namespace. Perhaps it could go under logger, like namespace logger { void Setup(...) }

Server/shared/logger.cpp
```diff
@@ -0,0 +1,86 @@
...
+	// setup file logger
+	std::string defaultLogPath = std::format("logs/{}.log", name);
+	std::string fileName = ini.GetString(ini::LOGGER, ini::FILE, defaultLogPath);
+	auto fileLogger = std::make_shared<spdlog::sinks::daily_file_format_sink_mt>(fileName, 0, 0);
+
+	// setup console logger
+	auto consoleLogger = std::make_shared<spdlog::sinks::stdout_color_sink_mt>();
+
+	// setup multi-sink async logger as default (combines file+console logger)
+	spdlog::init_thread_pool(8192, 1);
```
twostars: These magic numbers should be constants

Server/shared/logger.cpp
```diff
@@ -0,0 +1,86 @@
...
+	std::vector<spdlog::sink_ptr> sinks{fileLogger, consoleLogger};
+	std::shared_ptr<spdlog::details::thread_pool> threadPool = spdlog::thread_pool();
```
twostars: These convoluted types can just be auto

All.sln
```diff
@@ -174,6 +174,13 @@ Project("{8BC9CEB8-8B4A-11D0-8D11-00A0C91BC942}") = "db-library", "deps\db-libra
 		{057C1ABC-6137-37A9-9935-8CF811D7CEDA} = {057C1ABC-6137-37A9-9935-8CF811D7CEDA}
 	EndProjectSection
 EndProject
+Project("{8BC9CEB8-8B4A-11D0-8D11-00A0C91BC942}") = "fetch-spdlog", "deps\fetch-and-build-wrappers\fetch-spdlog.vcxproj", "{10898C84-AF45-F0B0-5144-EA801DF15503}"
```
twostars, line 182: This applies to both solutions (All.sln and Server.sln). This line isn't really the issue, so ignore that.
The dependencies here are kind've ... not setup properly.

I'm sure you were just including the spdlog project for testing, which is fine. It should just be removed from the solution now though.

It, however, is the only thing depending on fetch-spdlog. The projects that need it should depend on it. I think ultimately shared (which the server projects all depend on), and db-library? I think that's all that needs to depend on it offhandedly.

twostars (changes requested):

Server/shared/logger.cpp
```diff
@@ -0,0 +1,86 @@
...
+#include <spdlog/common.h>
+#include <spdlog/sinks/stdout_color_sinks.h>
+#include <spdlog/async.h>
+#include <spdlog/async_logger.h>
+
+/// \brief Sets up spdlog from an ini file using standardized server settings
+/// \param ini server application's ini file (already loaded)
+/// \param name application name (VersionManager, Aujard, AIServer, Ebenezer)
+void SetupLogger(CIni& ini, const std::string& name)
+{
```
twostars, line 16: As far as I can tell, spdlog isn't ever shutdown gracefully via spdlog::shutdown()

twostars (changes requested):

twostars (approved):

PR #346, opened 2025-07-31
Store logs under the executable's path

twostars:
Everything else is loaded relative to the executable, so we should be sure to store logs there too. Otherwise they end up in the working dir which won't be the same when running under a debugger with default config.

### DIFF
--- a/Server/shared/logger.cpp
+++ b/Server/shared/logger.cpp
@@ -9,43 +9,53 @@
 #include <spdlog/async.h>
 #include <spdlog/async_logger.h>
 
+namespace logger
+{
+	void SetupExtraLogger(CIni& ini,
+		const std::string& appName, const std::string& logFileConfigProp,
+		const std::string& exePath,
+		std::shared_ptr<spdlog::details::thread_pool> threadPool);
+}
+
 /// \brief Sets up spdlog from an ini file using standardized server settings
 /// \param ini server application's ini file (already loaded)
-/// \param name application name (VersionManager, Aujard, AIServer, Ebenezer)
-void logger::SetupLogger(CIni& ini, const std::string& name)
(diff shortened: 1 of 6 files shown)


Server/shared/logger.cpp
```diff
@@ -9,74 +9,86 @@
...
 /// \param ini server application's ini file (already loaded)
-/// \param name application name (VersionManager, Aujard, AIServer, Ebenezer)
-void logger::SetupLogger(CIni& ini, const std::string& name)
+/// \param appName application name (VersionManager, Aujard, AIServer, Ebenezer)
+/// \param baseDir base directory to store logs folder under. must end in a trailing slash.
+void logger::SetupLogger(CIni& ini, const std::string& appName, const std::string& baseDir)
 {
 	// setup file logger
-	std::string defaultLogPath = std::format("logs/{}.log", name);
+	std::string defaultLogPath = std::format("logs/{}.log", appName);
```
kenner2, line 27: I think you want to put your baseDir here so that it defaults to an absolute path but allows for the user to override the config to be where-ever they want it to go.

Server/shared/logger.cpp
```diff
@@ -9,74 +9,86 @@
...
 
-void logger::SetupExtraLogger(CIni& ini, const std::string& appName,
-	const std::string& logFileConfigProp,
+void logger::SetupExtraLogger(CIni& ini,
+	const std::string& appName, const std::string& logFileConfigProp,
+	const std::string& baseDir,
 	std::shared_ptr<spdlog::details::thread_pool> threadPool)
 {
 	// setup file logger
 	std::string defaultLogPath = std::format("logs/{}.log", appName);
```
kenner2, line 75: same here

PR #360, opened 2025-08-03
Remove/translate korean in server .rc resource files

kenner2:
## Pull request type

The resource files are a mix of Korean/English.

## What is the new behaviour?
The resource files will only use English.

More usable UI. Less UTF-16 characters. Text resources in ebenezer were updated using SERVER_RESOURCE values.

## Demo
image

## Checklist

### DIFF
--- a/Server/AIServer/Server.rc
+++ b/Server/AIServer/Server.rc
@@ -19,6 +19,44 @@
 LANGUAGE LANG_ENGLISH, SUBLANG_NEUTRAL
 #pragma code_page(1252)
 
+#ifdef APSTUDIO_INVOKED
+/////////////////////////////////////////////////////////////////////////////
+//
+// TEXTINCLUDE
+//
+
+1 TEXTINCLUDE 
+BEGIN
+    "resource.h\0"
+END
+
+2 TEXTINCLUDE 
+BEGIN
+    "#include ""afxres.h""\r\n"
+    "\0"
+END
+
+3 TEXTINCLUDE 
+BEGIN
+    "#define _AFX_NO_SPLITTER_RESOURCES\r\n"
+    "#define _AFX_NO_OLE_RESOURCES\r\n"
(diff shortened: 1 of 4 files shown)


Server/AIServer/Server.rc
```diff
@@ -13,32 +13,30 @@
 #undef APSTUDIO_READONLY_SYMBOLS
 
 /////////////////////////////////////////////////////////////////////////////
-// Korean resources
+// Korean (Korea) resources
 
 #if !defined(AFX_RESOURCE_DLL) || defined(AFX_TARG_KOR)
-#ifdef _WIN32
```
kenner2, line 19: These files were updated through VS's UIs.  I don't think this being removed is an issue but I'll point it out

Server/AIServer/Server.rc
```diff
@@ -156,27 +156,47 @@ BEGIN
...
-    IDS_ELMORAD_CATCH_2     "������ �� 2 ����� ���ɵǾ����ϴ�"
-    IDS_KARUS_PATHWAY       "***  ī�罺�� ���� ��ΰ� ���Ƚ��ϴ�   ***"
-    IDS_ELMORAD_PATHWAY     "***  ������� ���� ��ΰ� ���Ƚ��ϴ�  ***"
+    IDS_ABOUTBOX            "Server information (&A)"
+    IDS_KARUS_CATCH_1       "Karus' first fort was captured."
+    IDS_KARUS_CATCH_2       "Karus' second fort was captured."
+    IDS_ELMORAD_CATCH_1     "Elmorad's first fort was captured."
+    IDS_ELMORAD_CATCH_2     "Elmorad's second fort was captured."
+    IDS_KARUS_PATHWAY       "*** The path to Karus has been opened. ***"
+    IDS_ELMORAD_PATHWAY     "*** The path to Elmorad has been opened. ***"
```
kenner2, line 186: None of these were in SERVER_RESOURCE

Issue #271, opened 2025-07-03
Memory leak???


UTengine:
When you create a warrior character or have one I'm not sure if it was elmorad or karus. Once you reach character select you drop massively in FPS. And also when you just finished loading into the game, the very first moments it stutters and has issues before it recovers, this doesn't happen without them. I noticed this bug a while ago and it happens when overwriting the chrselect and char files from 1264-1298. 1097 files seem to be fine.
This is not only related to OpenKO, also other projects I've worked on in the past.
Sorry for opening the issue just now.
Either there is something wrong with that particular file or our client doesn't handle it properly. Not sure if it's mesh related or animation related all I can state is this is the outcome.
@twostars

UTengine:
Yeah easily I believe its the chr directory lol the 1097 one fixes it. And it only happens once tou have a barbarian I think pinpointing it will be very easy just need to be able to run the game. Ill try to do asap

koraykarakus:
leaks caused by new usage can easily be tracked by implementing new macro for debug mode. I have tested before, none of the leaks caused by new / forgetting to delete, there are some leaks but probably because of other problems than new / delete. I never tested select character though

twostars:
There's no evidence of the aforementioned problem being a memory leak in the first place. From the problem description it's possible it's something like a file incorrectly loading, causing unintentionally excessive allocations from bad data and having it tick more during runtime as a result, but even that's a bit of a stretch.

Point is, the problem as described doesn't really sound like a leak, so assuming it's one is probably counter-intuitive. Better off just tracking down which file actually causes this and going from there, since it seems pretty reproducible from what's been said.

Best not to assume things and act based on that. Debug based on what you actually know, not what you assume to be the case.

twostars:
@UTengine will you be able to isolate this soon?

UTengine:
Sure I loaded yesterday and had 2 chars and a katus barbarian and it still happens btw

twostars:
>  I noticed this bug a while ago and it happens when overwriting the chrselect and char files from 1264-1298.
> 1097 files seem to be fine.

Can you test selectively overwrite these files to isolate it to the exact file causing the problem?

UTengine:
I'll try fucking laptop screen won't turn on I'll hard reset it omw

UTengine:
https://github.com/user-attachments/assets/e9ad5dbe-56d5-4863-8479-8b86dafa68d1

Is something looping?

UTengine:
[...]

PR #362, opened 2025-08-03
Charfps

UTengine:
CN3Base::s_SndMgr.ReleaseObj(&m_pSnd_Rotate);
m_pSnd_Rotate =CN3Base::s_SndMgr.CreateObj(ID_SOUND_CHR_SELECT_ROTATE); seems to be causing massive fps drops
a dirty move to make it instantly play and stop seems to fix the issue. temporary fix

- [x] Bugfix
- [x] Other (please describe):

https://github.com/Open-KO/KnightOnline/issues/271

## What is the new behaviour?
Just a play stop instantly might not be the perfect way, but having the game drain everything is worse.

Something was hugging up and it would only resolve once you rotated the chairs I tried initing lightning, animations, camera positions nothing helped untill i removed sounds from it then it was fine. Instead of going through all the hassle i thought about playing it once that did the trick but it played the ugly sound so I just did this instead.

## Demo
https://github.com/Open-KO/KnightOnline/issues/271

## Checklist

### DIFF
--- a/Client/WarFare/GameProcCharacterSelect.cpp
+++ b/Client/WarFare/GameProcCharacterSelect.cpp
@@ -218,7 +218,10 @@ void CGameProcCharacterSelect::Init()
 	if (s_bIsRestarting)
 		MsgSend_VersionCheck();
 	else
+	{
 		MsgSend_RequestAllCharacterInfo(); // 캐릭터 정보 요청..
+		m_pSnd_Rotate->Play(); m_pSnd_Rotate->Stop();
+	}
 }
 
 void CGameProcCharacterSelect::Tick()
--- a/Client/WarFare/GameProcCharacterSelect.cpp
+++ b/Client/WarFare/GameProcCharacterSelect.cpp
@@ -107,6 +107,7 @@ void CGameProcCharacterSelect::Init()
 
 	CN3Base::s_SndMgr.ReleaseObj(&m_pSnd_Rotate);
 	m_pSnd_Rotate = CN3Base::s_SndMgr.CreateObj(ID_SOUND_CHR_SELECT_ROTATE);
+	m_pSnd_Rotate->Play(); m_pSnd_Rotate->Stop();
 	s_pUIMgr->EnableOperationSet(false); // 기존의 캐릭터 정보 패킷이 들어올때까지 UI 를 Disable 시킨다...
 
 	m_pCamera = new CN3Camera();
@@ -218,10 +219,7 @@ void CGameProcCharacterSelect::Init()
 	if (s_bIsRestarting)
 		MsgSend_VersionCheck();
 	else
-	{
 		MsgSend_RequestAllCharacterInfo(); // 캐릭터 정보 요청..
-		m_pSnd_Rotate->Play(); m_pSnd_Rotate->Stop();
-	}
 }
 
 void CGameProcCharacterSelect::Tick()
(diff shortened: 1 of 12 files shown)


Client/WarFare/GameProcCharacterSelect.cpp
```diff
@@ -107,6 +107,7 @@ void CGameProcCharacterSelect::Init()
 
 	CN3Base::s_SndMgr.ReleaseObj(&m_pSnd_Rotate);
 	m_pSnd_Rotate = CN3Base::s_SndMgr.CreateObj(ID_SOUND_CHR_SELECT_ROTATE);
+	m_pSnd_Rotate->Play(); m_pSnd_Rotate->Stop();
```
kenner2, line 110: Put each line of code on its own line, and please surround with a comment block explaining the issue with a TODO to come back and remove this.  If you understand the nature of the issue, please open an issue stating what needs to be looked at and link to it as part of the TODO comment.

UTengine: It will just clutter up more.
The issue is linked, and twostars is going to investigate why this part is causing so much disaster.
What I did here was a quick fix and temporary holder I'll quote what he said:

> **twostars** —
> I need to take a good look at your issue with the sound thing
> see why exactly that is
> if it's a problem with all sounds or specific to that sound
> based on your fix it appears that preloading it stops the issue, but what does that mean for all other sounds in the game
> they all load miserably slow?
> hm
> lol
> Tahsin — 16:08
> Thats what I thought too you know
> But then questioned myself maybe its just related that this one is in a diff procedure and game proc ticks while its outside of it creating some big buffer or something or hogging down cycles. Because once you get ingame as you see in the footage when you get to walk a bit then it clears out. So i think its hogging the queue deque. Untill its/pointer released
> **twostars** — 16:11
> I mean it's only called when you first invoke it, it isn't like it's constantly calling Play() every frame or anything, and all you're doing is causing it to load once and stop it
> and apparently that fixes the issue
> so
> presumably something is loading very poorly there
> question is whether or not it's something specific to that file
> Tahsin — 16:12
> Yeah but when you create the object it gets added to a list right
> **twostars** — 16:12
> that audio file could just suck
> Tahsin — 16:12
> And the system queues deques afaik
> **twostars** — 16:12
> yeah but that's irrelevant because you're still just creating it once
> with your changes it still plays exactly as it did
> it still allocates exactly as it did
> you just invoke a play and tell it to stop
> which presumably forces it to load the sound file into memory
> which is presumably very slow
> **twostars** — 16:15
> I'm assuming just based on your proposed fix that it only loads on the initial play
> I haven't really touched the original audio
> I always replaced it fairly early on
> Tahsin — 16:15
> Thats what i think too tbh
> **twostars** — 16:15
> so I'd need to look into exactly what it's doing
> but yeah the bigger question is
> is that audio file jsut broken in some way
> or is it just very slow for them all
> Tahsin — 16:16
> But how its just a wav file and not that long tbh
> **twostars** — 16:16
> stuff to check
> doesn't mean there can't be bugs loading it

Other than that you're right, but the PR is on hold anyway, I won't change it as it will show multiple entries which probably won't be needed, and if we do go this way I will do as you told me. Github desktop isn't that friendly :'(

twostars:
Closed due to implementation via #365 and #370.

PR #365, opened 2025-08-05
One-time decode MP3 -> WAV using mpg123

### LINKED ISSUE #218: Update audio engine
I thought there was an issue for this somewhere but I couldn't find it.
Regardless, the audio engine needs to be updated and we need to decide where we want to go with it.

In my mind, there are 2 main choices here:
1. 1.298 officially uses the proprietary Miles Sound System for MP3 support. This can be hooked up, but -
2. More modern versions swap to libogg/libvorbisfile which is open-source, and swap their assets to .OGG.
This seems like the far more practical and forward-thinking option. The downside is it deviates from official (1.298) behaviour, and forces us to also update our assets, but it also isn't really unofficial since they ended up updating to it officially anyway.

twostars:
This PR introduces mpg123 (https://github.com/madebr/mpg123), currently built with no assembly optimisations (the compiler should do a decent enough job, though). For those we'd need to hook up yasm (which is old and tedious), but performance isn't really much of an issue here.

When loading sounds (either streamed or otherwise), we detect if they're mp3 via their file extension and attempt to decode to WAV, using mpg123.
We output these to \<filename\>.mp3.wav in order to identify converted files, so they can be ignored from the repo and so we don't assume any existing old WAV files match the newer MP3s.

If this file exists already, we use it directly, avoiding future conversions.

We force the client to run with adimn privs; these are needed to write the converted files out, which isn't really something specifically needed for this. The client needs admin privs officially *anyway* for matching registry access and consistent access for files like logs, so it's just something that will have needed to be swapped over at some point anyway.

Notably I pass the filenames through by value now, forcing copies. This isn't exactly ideal but a copy is needed any way I slice it (you could almost get away with a string_view if not for the map access, which needs to transform back to a string anyway), so it's just the simplest option, short of explicitly copying it.

As far as performance goes, it's not terrible. Considering there's also a grand total of only 9 MP3s, I don't think space or performance is a huge concern - very few files to actually convert, and they'd only be done once upon first encountering them, and never again.

I was considering just loading it directly up into memory, but that would require modifying two separate interfaces for it.
This just feels like the least obtrusive method. Considering with how very few MP3s there even are, I don't think it really matters.

Note: I'm aware I didn't touch existing background music behaviour for the 1.298 scene. This should be done in a separate PR.

Closes #218


Client/WarFare/WarFare.vcxproj
```diff
@@ -59,6 +59,7 @@
       <SubSystem>Windows</SubSystem>
       <AdditionalDependencies>shared.lib;ws2_32.lib;winmm.lib;imm32.lib;wsock32.lib;DInput8.lib;N3Base_client.lib;%(AdditionalDependencies)</AdditionalDependencies>
       <AdditionalLibraryDirectories>..\..\Server\shared\$(Configuration);%(AdditionalLibraryDirectories)</AdditionalLibraryDirectories>
+      <UACExecutionLevel>RequireAdministrator</UACExecutionLevel>
```
kenner2, line 62: Does writing the .mp3.wav file require elevation, or is it something else?

twostars: When installed, it typically does under Program Files. Or just anywhere that isn't directly user-writeable.
In the PR description:

> We force the client to run with admin privs; these are needed to write the converted files out, which isn't really something specifically needed for this. The client needs admin privs officially anyway for matching registry access and consistent access for files like logs, so it's just something that will have needed to be swapped over at some point anyway.

PR #367, opened 2025-08-05
implementation of UIMsgBoxOkCancel

koraykarakus:
## Pull request type

- [x] Feature

Currently in transactions UI confirm box feature is missing. Officially, upgrade items, unique items and scrolls should be sellable after confirmation.

## What is the new behaviour?

UIMsgBoxOkCancel is added to provide confirm dialog before selling mentioned items.

This is core feature which prevents user from selling important items ( & maybe buying expensive ones from NPC ).

Buying part is unclear because I could not test it with Pet items on original client because those items demand clan grade, and that does not work well.

Had problems while setting Z order of newly added UI and so I took CountableItemEdit class as an example while writing the code.

## Demo
image

## Checklist

### DIFF
--- a/Client/WarFare/GameDef.h
+++ b/Client/WarFare/GameDef.h
@@ -292,7 +292,7 @@ enum e_ItemClass	{	ITEM_CLASS_DAGGER = 11, // dagger
 						ITEM_CLASS_ARMOR_PRIEST = 240, // Priest armor
 
 						ITEM_CLASS_ETC = 251, // Miscellaneous
-
+						ITEM_CLASS_POWER_SCROLL = 255, //scrolls,premiums and power up potions, pets
 						ITEM_CLASS_UNKNOWN = 0xffffffff }; // 
 
 enum e_Nation { NATION_NOTSELECTED = 0, NATION_KARUS, NATION_ELMORAD, NATION_UNKNOWN = 0xffffffff };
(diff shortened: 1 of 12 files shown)


twostars:
I'll submit a proper review eventually, but for now I'll just note one major issue I have with this PR:

The problem with making it static (ignoring the fact it's named m\_\*, indicating a member var, not a static var -- and I know the other ones there are named the same, but they shouldn't be), is how this will interact with other UIs that need to use it.

So I checked over all the UIs that use it officially, and they are:
CUIDisguise, CUINationSelect, CUIPartyBBS, CUIPerTradeDlg, CUITransactionDlg, CUIWarfareRequestList

From here, the question was -- of these, are they all mutually exclusive? Disregarding CUINationSelect, since it's not an ingame UI, the answer is *mostly* yes EXCEPT for CUIPartyBBS.

This one can show up concurrently with other UIs, which means it cannot be static. It needs to be instanced per UI that requires it.
Instancing it will also avoid issues with things like this *very* iffy logic, which can very easily cause unintentional bugs depending on resulting order of operation with UIs:
```cpp
void CUIMsgBoxOkCancel::SetVisibleWithNoSound(bool bVisible, bool bWork, bool bReFocus)
{
	if (bWork)
	{
		ReceiveMessage(m_pBtn_Cancel, UIMSG_BUTTON_CLICK);
	}
```

Instancing it will also solve your problem of z-order, since it'll simply be a child of the UI you load it in, and not its own independent UI.

PR #382, opened 2025-08-07
Load CKnightsManager strings from Win32 resources

koraykarakus:
## Pull request type

- [ ] Bugfix
- [ ] Feature
- [ ] Build related changes

Some strings are not inside RC file and they are in Korean language.

## What is the new behaviour?

Translated to English and placed inside rc file.

Note: Some of the translated texts might be wrong

## Checklist


koraykarakus:
added that table to RC and corrected all the changed IDs but if you plan to chose from DB and use Find function with ID, I can understand.

twostars:
We will load the table and pull them from there, but for now at least the resource definitions are in, as well as the handful of calls you replaced for swapping out there.

The resources themselves will get stripped once that table's loaded though.

Issue #386, opened 2025-08-08
Sell prices of some items displayed wrong on tooltip ( Gold bar, silver bar, coins )


koraykarakus:
Sell prices of mentioned items displayed wrong on tooltip.

PR #390, opened 2025-08-08
fixing image tooltip for saletype = 1 items[gold bar, silver coin etc.]

koraykarakus:
## Pull request type

- [x] Bugfix

Tooltip prices are not updated with respect to sale type, related to issue #386 

## What is the new behaviour?

Sale type 1,0 and 2 handled accordingly. Premium features should be tested and implemented if required.

iPriceSale was misleading and renamed under __TABLE_BASICS. UITooltip has been updated to provide correct results.

## Checklist

### DIFF
--- a/Client/WarFare/UIImageTooltipDlg.cpp
+++ b/Client/WarFare/UIImageTooltipDlg.cpp
@@ -1164,7 +1164,18 @@ exceptions:;
 			}
 			else
 			{	
-				int iSellPrice = (spItem->pItemBasic->iPrice*spItem->pItemExt->siPriceMultiply/6);
+				int iSellPrice = 0;
+
+				if (spItem->pItemBasic->iSaleType == 1) 
+				{   //sale type = 1, selling for the same price of buying
+					iSellPrice = spItem->pItemBasic->iPrice * spItem->pItemExt->siPriceMultiply;
+				}
+				else
+				{
+					//non premium : selling for 1/6 of purchasing price
+					iSellPrice = spItem->pItemBasic->iPrice * spItem->pItemExt->siPriceMultiply / 6;
+				}
+
 				if (iSellPrice < 1)
 					iSellPrice = 1;
 
--- a/Client/WarFare/GameDef.h
+++ b/Client/WarFare/GameDef.h
@@ -706,7 +706,7 @@ struct __TABLE_ITEM_BASIC
 	int16_t		siWeight;				// 19 Weight (in 0.1 units)
 	int16_t		siMaxDurability;		// 20 Max durability
 	int			iPrice;					// 21 Purchase price
-	int			iPriceSale;				// 22 Sale price
+	int			iSaleType;				// 22 Sale type ([1 - same price], [0 - 1/6 of buy price], [2 - 1/6 & cant be repaired] ) - [1/4 premium(all)]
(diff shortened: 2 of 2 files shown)


Client/WarFare/UIImageTooltipDlg.cpp
```diff
@@ -1065,7 +1065,18 @@ exceptions:;
 			}
 			else
 			{	
-				int iSellPrice = (spItem->pItemBasic->iPrice*spItem->pItemExt->siPriceMultiply/6);
+				int iSellPrice = 0;
+
+				if (spItem->pItemBasic->iSaleType == 1)
```
twostars: This should use an enum, rather than magic numbers.

Client/WarFare/UIImageTooltipDlg.cpp
```diff
@@ -1065,7 +1065,18 @@ exceptions:;
 			}
 			else
 			{	
-				int iSellPrice = (spItem->pItemBasic->iPrice*spItem->pItemExt->siPriceMultiply/6);
+				int iSellPrice = 0;
+
+				if (spItem->pItemBasic->iSaleType == 1) 
+				{   //sale type = 1, selling for the same price of buying
```
twostars: With an enum making this clear, this comment isn't necessary, but your comments should always have a leading space, i.e.:
`// comment`
(it also shouldn't be on {)

Client/WarFare/UIImageTooltipDlg.cpp
```diff
@@ -1065,7 +1065,18 @@ exceptions:;
...
+					iSellPrice = spItem->pItemBasic->iPrice * spItem->pItemExt->siPriceMultiply;
+				}
+				else
+				{
+					//non premium : selling for 1/6 of purchasing price
```
twostars: Your comments should always have a leading space, i.e.:
`// comment`

As far as the premium-related logic, this behaviour extends further than just here. Rather than specifically remark upon it everywhere, it would be better to just have it call a method to apply it as needed, e.g. `__IconItemSkill::GetSellPrice()`.

Then you wouldn't have to constantly explain why it's divided by 6 here; you'd just handle it the once and reuse it.

twostars (changes requested):

Client/WarFare/UIImageTooltipDlg.cpp
koraykarakus: Maybe we should go with instance, spItem already presents in code, so we can get as spItem->GetSellPrice(bHasPremium)

instead of static which requires resending pointers as follows : 
__IconItemSkill::GetSellPrice(__TABLE_BASIC* pItemBasic, __TABLE_EXT* pItemExt, bHasPremium)

In addition to this, GetBuyPrice() should be implemented, I ll commit those now but I want to keep premium feature for later. Premium requires testing, I am unsure if premium prices divided on tooltip of original exe. To test premium features efficiently GM commands also needed in my case, right now working on the project with extreamly low RAM, when everything is open, MSSQL works too slow and querying becomes hard.

koraykarakus: 1 final requested change is not visible github bug.

twostars (approved):

Issue #397, opened 2025-08-09
UIInventory, imagetooltip should be updated if changing item is successfull


koraykarakus:
After right clicking on item to equip it, - if player does not move mouse - tooltip displays wrong item's information. It is updated on original version after successfull equip.

PR #399, opened 2025-08-10
Ensure item tooltips update on item change

koraykarakus:
## Pull request type

- [x] Bugfix

Image tooltip info does not refresh when an Item equiped successfully if the cursor stays at the same location.

## What is the new behaviour?

Tooltip refresh after equipping item.

Setter function added to UIImageTooltip to externally change its Rewrite state. Related to issue #397 

## Checklist

### DIFF
--- a/Client/WarFare/UIImageTooltipDlg.cpp
+++ b/Client/WarFare/UIImageTooltipDlg.cpp
@@ -210,7 +210,7 @@ int	CUIImageTooltipDlg::CalcTooltipStringNumAndWrite(__IconItemSkill* spItem, bo
 
 			return iIndex;	
 		}
-		else if ( spItem->pItemBasic->szName != m_pStr[iIndex]->GetString() )
+		else
 		{
 			e_ItemAttrib eTA = (e_ItemAttrib)(spItem->pItemExt->byMagicOrRare);
 			switch (eTA)


koraykarakus:
When equipping an item from the inventory to the arms, the inventory icon updates, but the tooltip still displays the previously equipped item as long as the cursor remains stationary. Once the cursor is moved, the tooltip updates with the new item’s information.

Currently, UIImageTooltip uses position checks and avoids refreshing the text while the cursor is stationary for efficiency. This makes sense for the same icon and same item, since the data does not change and only the tooltip position needs updating. However, there are cases where the tooltip data must be refreshed even if the cursor is not moved. To handle this, a state check has been implemented. For future flexibility, a setter function and a private member were added, so that refreshing can be triggered via the setter instead of directly accessing members like m_pUITooltip->m_needRefresh = true; from other classes.

This behavior also exists in the official client. The issue could be fixed by removing the same-location check entirely, but the original developer likely kept it to skip unnecessary calculations when the cursor is stationary.

For other UI elements that use the same tooltip system, if the icon changes while the cursor is stationary, they would also require an update. However, confirming this would require testing and reproducing the issue in those cases.

koraykarakus:
yes, it is understandable now, i did not investigate code in detail 

if ( (!m_spItemBack) || (m_spItemBack->pItemBasic->dwID != spItem->pItemBasic->dwID) || 
		(m_spItemBack->pItemExt->dwID != spItem->pItemExt->dwID) ||
		(m_spItemBack->iDurability != spItem->iDurability) )
	{
	
just skipped reading that by thinking some controlls i ll refine PR when possible

twostars:
Note that behaviour is different now - you're updating the entire tooltip when the cursor moves. Item caching would previously kick in before and prevent the update.

The tooltip itself should only update when the item changes.
If the item hasn't changed but the cursor has moved, it should still be repositioned (to track the cursor).

PR #416, opened 2025-08-13
Allow level 80 characters to login

koraykarakus:
## Pull request type

- [x] Bugfix

cannot login game if level is 80

## What is the new behaviour?

can login now.

important for testing 80 level skills.

## Checklist

### DIFF
--- a/Server/Ebenezer/User.cpp
+++ b/Server/Ebenezer/User.cpp
@@ -2609,13 +2609,17 @@ void CUser::SendTimeStatus()
 
 void CUser::SetDetailData()
 {
-	C3DMap* pMap = nullptr;
-
 	SetSlotItemValue();
 	SetUserAbility();
 
 	if (m_pUserData->m_bLevel > MAX_LEVEL)
+	{
+		spdlog::error("User::SetDetailData: user exceeds max level [accountId={} charId={} level={}]",
+			m_pUserData->m_Accountid, m_pUserData->m_id, m_pUserData->m_bLevel);
+
 		Close();
+		return;
+	}
 
 	m_iMaxExp = m_pMain->m_LevelUpTableArray[m_pUserData->m_bLevel - 1]->RequiredExp;
 	m_iMaxWeight = (m_pUserData->m_bStr + m_sItemStr) * 50;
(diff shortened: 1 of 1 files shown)


twostars:
This fixes the DC, but it should we should ideally return here too, to avoid it bothering to access the container (not that it's directly relevant to the container lookup, but the code's intent is to prevent it from proceeding further):

```cpp
	if (m_pUserData->m_bLevel > MAX_LEVEL)
		Close();

	m_iMaxExp = m_pMain->m_LevelUpTableArray[m_pUserData->m_bLevel - 1]->RequiredExp; // <--
```

Server/Ebenezer/User.cpp
```diff
@@ -2614,7 +2614,7 @@ void CUser::SetDetailData()
 	SetSlotItemValue();
 	SetUserAbility();
 
-	if (m_pUserData->m_bLevel >= MAX_LEVEL)
+	if (m_pUserData->m_bLevel > MAX_LEVEL)
```
kenner2, line 2615: Add braces to this block so it can support multi-line and add a logger line:
spdlog::error("User::SetDetailData: user exceeds max level [accountId={} charId={} level={}]", m_pUserData->m_strAccountId, m_pUserData->m_id, m_pUserData->m_bLevel);

Variable names might be off - just pulling from memory

koraykarakus: good idea to log it on official project but why not just set user level equal to MAX_LEVEL on the condition user exceeded that ? Stats and skills also be set accordingly, otherwise owner of server should deal with logcheck or ticket answering.

An user shall not exceed that limit initially, but imaginary situation in which server level limits changed.

twostars: If such a case occurred, then that's on the server owner to fix. It shouldn't happen unless things are misconfigured or there's a bug causing their level to be higher than intended.
We shouldn't just fix the level and stats/skill points here because we shouldn't assume what happened. It's an exceptional case.

The limit could probably be enforced more consistently (though as-is, official behaviour requires extra level[s] above cap), rather than using the hardcoded limit, but it's just wonky official behaviour that needs a bit of thought to get right. It'd probably use a combination of config (INI) + DB checks to get right (since some of them are intended to just prevent out-of-bounds reads from m_LevelUpArray).

Regardless, better to just leave it and not make assumptions.

PR #419, opened 2025-08-13
removal of irrelevant check

koraykarakus:
## Pull request type

- [x] Bugfix

Having irrelevant code inside SkillMng.cpp can cause future problems.

## What is the new behaviour?

Irrelevant controller removed.

This part of code checks for items which has byAttachPoint = 17 & byNeedRace = playerRace
there are no such items, byNeedRace probably implemented / translated wrong. only a few items have byNeedRace rather than 0 and those items' value is 20. None of those items with 20 value has byAttachPoint = 17, they differ but never 17. Those items are dumplings (user kill rewards), rice cakes etc.

## Checklist


koraykarakus:
ITEM_LIMITED_EXHAUST that is also needed in server side. That block checks if an user can use class limited items for example priest book from sundries.

As far as I saw, server side does not have check for it.

PR #420, opened 2025-08-14
remove type 4 magic check from client

koraykarakus:
## Pull request type

- [x] Bugfix

magicskillmanager on client checking for type 4 skills and preventing sendin packet. this causes unofficial behavior ( character should act and try to buff himself and get skill failed )

## What is the new behaviour?

works as official.

tested orginal client with openko server. orginal client sends packet even if user has the buff. there is no client filter for type 4 magic ( can be tested with priest buffs, used grace ).

## Checklist

### DIFF
--- a/Client/WarFare/MagicSkillMng.cpp
+++ b/Client/WarFare/MagicSkillMng.cpp
@@ -854,37 +854,7 @@ bool CMagicSkillMng::CheckValidCondition(int iTargetID, __TABLE_UPC_SKILL* pSkil
 		}
 	}
 
-	if( (pSkill->dw1stTableType==4 || pSkill->dw2ndTableType==4) && 
-		( (pSkill->iTarget==SKILLMAGIC_TARGET_SELF) || (iTargetID==s_pPlayer->IDNumber()) ) )
-	{
-		__TABLE_UPC_SKILL_TYPE_4* pType4 = m_pTbl_Type_4->Find(pSkill->dwID);
-		if(!pType4) return false;
-		
-		switch(pType4->iBuffType)
-		{
-		case BUFFTYPE_MAXHP:
-			if(m_iMaxHP != 0) return false;
-			break;
-		case BUFFTYPE_AC:
-			if(m_iAC != 0) return false;
-			break;
-		case BUFFTYPE_ATTACK:
-			if(m_iAttack != 0) return false;
-			break;
-		case BUFFTYPE_ATTACKSPEED:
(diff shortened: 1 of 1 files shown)


twostars:
image

koraykarakus:
The original client sends a packet on failure, even though the assembly conversion suggests otherwise. Replicating every detail and achieving identical behavior is not always possible. IDA provides only partial insights, and contributors to the project may differ in their ability to name, interpret, and resolve functions using IDA.

Opcode changes : as shown
C:\Users\Administrator\Desktop\openko_forked\KnightOnline\Server\Ebenezer\User.cpp(238) : atlTraceGeneral - Packet Length: 22 | Data: 31 03 F0 B7 01 00 00 00 00 00 00 00 00 00 00 00 00 00 00 00 00 00 
C:\Users\Administrator\Desktop\openko_forked\KnightOnline\Server\Ebenezer\User.cpp(238) : atlTraceGeneral - Packet Length: 22 | Data: 31 01 F0 B7 01 00 00 00 00 00 00 00 00 00 00 00 00 00 00 00 00 00
seems like op code changed from N3_SP_MAGIC_EFFECTING to N3_SP_MAGIC_CASTING

on the other hand type 3 does not send packet if ability is already active

twostars:
The point of my screenshot was to show that your conclusion about the behaviour you're seeing is clearly wrong, based on the fact it does still check officially.

And if your conclusion is wrong, then that puts your entire PR into question.

I would try to investigate what the source of your problem is first. If it can be thoroughly explained and the check still ends up useless, then by all means.
But I'm not going to just arbitrarily remove stuff based on conclusions you've made that I'm able to disprove.

koraykarakus:
It is really good you provide that information because reversing that with third party is huge time sink from scratch. Indeed we cannot remove function which actually exists, but it is required to implement additional code to imitate same behavior with seperate PR.

koraykarakus:
> Why does it need a separate PR? If you proof read what's in the code against the disassembled code in the screenshot you'll find there's a missing check that may be causing the issue.

it can fix additional bugs, but those skills not related to what I test ( priest buff grace skill I used to test ). 

this is the missing check:
(pSkill->dw1stTableType != 6 || pSkill->dw2ndTableType != 4) 
1st table 6 -> checks for transformation scroll's skills
2nd table 4 -> checks for slowing / stunning skills ( scream, cold wave, cold snap, leg cutting )

PR #427, opened 2025-08-15
correcting mana reduction for skill manager

koraykarakus:
## Pull request type

- [x] Bugfix

mana cost of a skill scales with total damage put.

## What is the new behaviour?

damage does not affecting mana cost.

Ebenezer/MagicProcess.cpp mana related parts edited.

## Checklist

### DIFF
--- a/Server/Ebenezer/MagicProcess.cpp
+++ b/Server/Ebenezer/MagicProcess.cpp
@@ -310,17 +310,14 @@ void CMagicProcess::MagicPacket(char* pBuf, int len)
 				// Only if Flying Effect is greater than 0.
 				if (pMagic->FlyingEffect > 0)
 				{
-					int total_hit = m_pSrcUser->m_sTotalHit + m_pSrcUser->m_sItemHit;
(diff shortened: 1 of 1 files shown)


Server/Ebenezer/MagicProcess.cpp
```diff
@@ -997,12 +983,10 @@ model::Magic* CMagicProcess::IsAvailable(int magicid, int tid, int sid, BYTE typ
 			}
 
 			// Actual deduction of Skill or Magic point.
-			if (pTable->Type1 == 1
-				|| pTable->Type1 == 2)
-				m_pSrcUser->MSpChange(-skill_mana);
-			else if (pTable->Type1 != 4
-				|| (pTable->Type1 == 4 && tid == -1))
+			if (pTable->Type1 != 4 || (pTable->Type1 == 4 && tid == -1))
```
twostars, line 986: This no longer applies to fallback cases for 1 & 2.

twostars (changes requested):

twostars (approved):

PR #435, opened 2025-08-16
Party Invite - Clean up party declines

reignsprime:
## Pull request type

- [x] Bugfix

Party invites can become bugged when there is a failed invite eg. other user declines or is outside of the permitted level range. This was causing the PartyUI to display and hide in unusual ways.

## What is the new behaviour?
Party invites work and the correct error codes/ logic is being triggered. This fixes the errors with the PartyUI. Also updated reference to iMSP as this Ext Info, not Base Info. Tested with multiple source clients connected to a binary server.

The ID and ErrorCode from the packet were being read in the wrong order, which was causing the logic to fail and unusual behaviour.

## Demo
N/A

## Checklist

### DIFF
--- a/Client/WarFare/GameProcMain.cpp
+++ b/Client/WarFare/GameProcMain.cpp
@@ -5097,11 +5097,11 @@ void CGameProcMain::MsgRecv_PartyOrForce(Packet& pkt)
 
 		case N3_SP_PARTY_OR_FORCE_INSERT:			// 0x03	// Send - s1(ID) | Recv - s3(ID, HPMax, HP, MPMax, MP) b2(Level, Class) - 문자열은 ID 로 알아낸다..
 		{
-			int iIDorErrorCode		= pkt.read<uint16_t>();		// IDs and positive numbers, error codes are negative numbers
-			int iPartyPosition		= pkt.read<int8_t>();		// order of user in the party
+			int iIDorErrorCode		= pkt.read<int16_t>();		// IDs and positive numbers, error codes are negative numbers
 
 			if (iIDorErrorCode >= 0)
 			{
+				int iPartyPosition	= pkt.read<uint8_t>();		// order of user in the party
 				int iIDLength		= pkt.read<int16_t>();
 				std::string szID;	pkt.readString(szID, iIDLength);
 				int iHPMax			= pkt.read<int16_t>();


Client/WarFare/GameProcMain.cpp
```diff
@@ -5097,8 +5097,8 @@ void CGameProcMain::MsgRecv_PartyOrForce(Packet& pkt)
 
 		case N3_SP_PARTY_OR_FORCE_INSERT:			// 0x03	// Send - s1(ID) | Recv - s3(ID, HPMax, HP, MPMax, MP) b2(Level, Class) - 문자열은 ID 로 알아낸다..
 		{
-			int iID = pkt.read<int16_t>();
-			int iErrorCode = pkt.read<uint8_t>();
+			int iErrorCode	= pkt.read<int8_t>();
```
twostars: This isn't actually correct. The problem is both the server (because it was reset) and client are inaccurate at this point, but the error code and user ID are actually a shared field.

When it's >= 0 (as in the below check), it acts as the user ID.
When it's negative, it acts as the error code.

Both the server and client need updating to behave this way for consistency with the official 1.298 client and server.

twostars (changes requested):

Client/WarFare/GameProcMain.cpp
reignsprime: Thanks, I will do some testing and work on it when I get a minute.

reignsprime: I did a bit of testing and determined you are correct. 

The binary server returns:

int iIDErrorCode = pkt.read<uint16_t>();   // If positive, it is the ID. If it is negative it is the error code
int iPartyPosition = pkt.read<int8_t>(); // The users position in the party ie. 0 is leader, 1 is the second all the way up to 7.

Client/WarFare/GameProcMain.cpp
```diff
@@ -5097,10 +5097,10 @@ void CGameProcMain::MsgRecv_PartyOrForce(Packet& pkt)
 
 		case N3_SP_PARTY_OR_FORCE_INSERT:			// 0x03	// Send - s1(ID) | Recv - s3(ID, HPMax, HP, MPMax, MP) b2(Level, Class) - 문자열은 ID 로 알아낸다..
 		{
-			int iID = pkt.read<int16_t>();
-			int iErrorCode = pkt.read<uint8_t>();
+			int iIDorErrorCode		= pkt.read<uint16_t>();		// IDs and positive numbers, error codes are negative numbers
```
twostars: This should be signed (int16_t not uint16_t).

Client/WarFare/GameProcMain.cpp
```diff
@@ -5097,10 +5097,10 @@ void CGameProcMain::MsgRecv_PartyOrForce(Packet& pkt)
...
+			int iPartyPosition		= pkt.read<int8_t>();		// order of user in the party
```
twostars: This only applies on successful packets; it's not included in error packets. This one's also unsigned, so it should be uint8_t.

twostars (approved):

PR #436, opened 2025-08-17
CUIKnights - Party Invite

reignsprime:
## Pull request type

- [x] Bugfix

Using the invite to party button in the clan window does not invite a user to a party.

## What is the new behaviour?
Clicking the invite button in the clan window invites the clan member to a party.

The function was missing.
Linked up the button from the UIF file.
Also introduced a check to prevent a user inviting themself to a party. 

## Demo
Button highlighted in yellow below.
KnightUI

## Checklist

### DIFF
--- a/Client/WarFare/UIVarious.cpp
+++ b/Client/WarFare/UIVarious.cpp
@@ -511,25 +511,25 @@ void CUIState::MsgSendAblityPointChange(uint8_t byType, int16_t siValueDelta)
 CUIKnights::CUIKnights()
 {
 	m_iPageCur = 1;
-	// m_MemberList = NULL;
+	// m_MemberList = nullptr;
 
-	m_pText_Name = NULL;
-	m_pText_Duty = NULL;
-	m_pText_Page = NULL;
-	m_pText_MemberCount = NULL;
+	m_pText_Name = nullptr;
+	m_pText_Duty = nullptr;
+	m_pText_Page = nullptr;
+	m_pText_MemberCount = nullptr;
 
-	// m_pImage_Grade = NULL;
+	// m_pImage_Grade = nullptr;
 
-	m_pList_CharGrades = NULL;
-	m_pList_CharIDs = NULL;
-	m_pList_CharLevels = NULL;
-	m_pList_CharJobs = NULL;
+	m_pList_CharGrades = nullptr;
+	m_pList_CharIDs = nullptr;
+	m_pList_CharLevels = nullptr;
+	m_pList_CharJobs = nullptr;
 
-	m_pBtn_Admit = NULL;
-	m_pBtn_Appoint = NULL;
-	m_pBtn_Remove = NULL;
-	m_pBtn_Refresh = NULL;
-	m_pBtn_ClanParty = NULL;
+	m_pBtn_Admit = nullptr;
+	m_pBtn_Appoint = nullptr;
+	m_pBtn_Remove = nullptr;
+	m_pBtn_Refresh = nullptr;
+	m_pBtn_ClanParty = nullptr;
 
 	m_fTimeLimit_Refresh = 0.0f;
 	m_fTimeLimit_Appoint = 0.0f;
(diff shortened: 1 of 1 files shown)


Client/WarFare/UIVarious.cpp
```diff
@@ -529,6 +529,7 @@ CUIKnights::CUIKnights()
 	m_pBtn_Appoint = NULL;
 	m_pBtn_Remove = NULL;
 	m_pBtn_Refresh = NULL;
+	m_pBtn_ClanParty = NULL;
```
twostars: Use nullptr instead of NULL; existing cases need to be updated, so we don't want to add to them.

Client/WarFare/UIVarious.cpp
```diff
@@ -594,20 +596,21 @@ bool CUIKnights::Load(HANDLE hFile)
 {
 	if (false == CN3UIBase::Load(hFile)) return false;
 
-	m_pText_Name = (CN3UIString*)this->GetChildByID("Text_ClansName");			__ASSERT(m_pText_Name, "NULL UI Component!!");
-	m_pText_Duty = (CN3UIString*)this->GetChildByID("Text_clan_Duty");			__ASSERT(m_pText_Duty, "NULL UI Component!!");
-	m_pText_Page = (CN3UIString*)this->GetChildByID("Text_clan_Page");			__ASSERT(m_pText_Page, "NULL UI Component!!");
+	m_pText_Name		= (CN3UIString*)this->GetChildByID("Text_ClansName");			__ASSERT(m_pText_Name, "NULL UI Component!!");
```
twostars: This should just be updated to N3_VERIFY_UI_COMPONENT(), i.e.:
```cpp
N3_VERIFY_UI_COMPONENT(m_pText_Name,  (CN3UIString*) GetChildByID("Text_ClansName"));
```

Client/WarFare/UIVarious.cpp
```diff
@@ -652,6 +655,8 @@ bool CUIKnights::ReceiveMessage(CN3UIBase* pSender, uint32_t dwMsg)
 		NextPageButtonHandler();
 	else if (pSender->m_szID == "btn_clan_refresh")
 		RefreshButtonHandler();
+	else if (pSender->m_szID == "btn_Clan_party")
```
twostars: These buttons are loaded. We don't need string comparisons here. We can just check the pointer, i.e.:
```cpp
else if (pSender == m_pBtn_ClanParty)
```

Client/WarFare/UIVarious.cpp
```diff
@@ -699,6 +704,34 @@ void CUIKnights::RefreshButtonHandler(bool blBypassTime)
 	MsgSend_MemberInfoAll();
 }
 
+void CUIKnights::ClanPartyButtonHandler()
+{
+	int iSel = m_pList_CharIDs->GetCurSel();
+	if (iSel == -1)
```
twostars: You can remove this check, and just check the result of `GetString()`.
It returns a bool depending on whether or not it's a valid index.

Client/WarFare/UIVarious.cpp
```diff
@@ -699,6 +704,34 @@ void CUIKnights::RefreshButtonHandler(bool blBypassTime)
...
+		return;
+
+	std::string szID;
+	m_pList_CharIDs->GetString(iSel, szID);
+
+	std::string myID = CGameProcedure::s_pPlayer->IDString();
```
twostars: 2 things:
1. We have no reason to copy this value. Prefer to use a const reference (since we're also not changing it), i.e.: `const std::string& myID = CGameBase::s_pPlayer->IDString();`
2. The other thing is that s_pPlayer is a `CGameBase` member, not `CGameProcedure`. It works because of inheritance, but it's incorrect and misleading.

Client/WarFare/UIVarious.cpp
```diff
@@ -699,6 +704,34 @@ void CUIKnights::RefreshButtonHandler(bool blBypassTime)
...
+	std::string szMsg;
+
+	// Prevent inviting yourself
+	if (szID == myID)
```
twostars: This needs to be a case-insensitive comparison, as it's dependent on the database data here, which doesn't tend to preserve the original character name's case, just how they were invited (AFAIK -- or the opposite, I don't remember). Officially they use `lstrcmpiA()` for these.

twostars (changes requested):

twostars (approved):

PR #440, opened 2025-08-17
UICmd : NULL to nullptr, N3_VERIFY to load UI

koraykarakus:
## Pull request type

NULL macro used and UI components loaded by old way.

## What is the new behaviour?

NULL converted to nullptr
N3_UI_VERIFY used to load components instead of __ASSERT

cleanup

## Checklist

### DIFF
--- a/Client/WarFare/UICmd.cpp
+++ b/Client/WarFare/UICmd.cpp
@@ -25,27 +25,27 @@ static char THIS_FILE[]=__FILE__;
 
 CUICmd::CUICmd()
 {
-	m_pBtn_Exit = NULL;			//나가기
+	m_pBtn_Exit = nullptr;			//나가기
 
-	m_pBtn_Act = NULL;			//행동
-	m_pBtn_Act_Walk = NULL;	//걷기
-	m_pBtn_Act_Run = NULL;		//달리기
-	m_pBtn_Act_Stop = NULL;	//서기
-	m_pBtn_Act_Attack = NULL;	//공격
+	m_pBtn_Act = nullptr;			//행동
+	m_pBtn_Act_Walk = nullptr;	//걷기
+	m_pBtn_Act_Run = nullptr;		//달리기
+	m_pBtn_Act_Stop = nullptr;	//서기
+	m_pBtn_Act_Attack = nullptr;	//공격
 	
-	m_pBtn_Act_StandUp = NULL; // 일어서기.
-	m_pBtn_Act_SitDown = NULL;	// 앉기
+	m_pBtn_Act_StandUp = nullptr; // 일어서기.
+	m_pBtn_Act_SitDown = nullptr;	// 앉기
 
-	m_pBtn_Camera = NULL;			//카메라
-	m_pBtn_Inventory = NULL;		//아이템 창 
-	m_pBtn_Party_Invite = NULL;	//파티 초대
-	m_pBtn_Party_Disband = NULL;	//파티 탈퇴
-	m_pBtn_CmdList = NULL;			//옵션
-	m_pBtn_Quest = NULL;			//퀘스트
-	m_pBtn_Character = NULL;		//자기 정보창   
-	m_pBtn_Skill = NULL;			//스킬트리 또는 마법창 
-	m_pBtn_Belong = NULL;			//소속 기사단 
-	m_pBtn_Map = NULL;				// 미니맵
+	m_pBtn_Camera = nullptr;			//카메라
+	m_pBtn_Inventory = nullptr;		//아이템 창 
+	m_pBtn_Party_Invite = nullptr;	//파티 초대
+	m_pBtn_Party_Disband = nullptr;	//파티 탈퇴
+	m_pBtn_CmdList = nullptr;			//옵션
+	m_pBtn_Quest = nullptr;			//퀘스트
+	m_pBtn_Character = nullptr;		//자기 정보창   
+	m_pBtn_Skill = nullptr;			//스킬트리 또는 마법창 
+	m_pBtn_Belong = nullptr;			//소속 기사단 
+	m_pBtn_Map = nullptr;				// 미니맵
 }
 
 CUICmd::~CUICmd()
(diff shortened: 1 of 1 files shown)


Client/WarFare/UICmd.cpp
```diff
@@ -56,31 +56,31 @@ CUICmd::~CUICmd()
...
-	m_pBtn_CmdList =			GetChildByID("btn_option");		//__ASSERT(m_pBtn_Option, "NULL UI Component!!");
-	m_pBtn_Camera =			GetChildByID("btn_camera");		//__ASSERT(m_pBtn_Camera, "NULL UI Component!!");
-	m_pBtn_Party_Invite =	GetChildByID("btn_invite");		//__ASSERT(m_pBtn_Party_Invite, "NULL UI Component!!");
-	m_pBtn_Party_Disband =	GetChildByID("btn_disband");	//__ASSERT(m_pBtn_Party_Disband, "NULL UI Component!!");
-	m_pBtn_Skill =			GetChildByID("btn_skill");		//__ASSERT(m_pBtn_Skill, "NULL UI Component!!");
-	m_pBtn_Exit	=			GetChildByID("btn_Exit");		//__ASSERT(m_pBtn_Exit, "NULL UI Component!!");
-
-	m_pBtn_Quest =		GetChildByID("btn_quest");			//__ASSERT(m_pBtn_Quest, "NULL UI Component!!");
-	m_pBtn_Belong =		GetChildByID("btn_knight");			//__ASSERT(m_pBtn_Belong, "NULL UI Component!!");
+	N3_VERIFY_UI_COMPONENT(m_pBtn_Act, GetChildByID("btn_control"));
```
twostars: Preserve indentation after the comma.

Client/WarFare/UICmd.cpp
```diff
@@ -56,31 +56,31 @@ CUICmd::~CUICmd()
...
+	N3_VERIFY_UI_COMPONENT(m_pBtn_Act_Walk, GetChildByID("btn_walk"));
+	N3_VERIFY_UI_COMPONENT(m_pBtn_Act_Run, GetChildByID("btn_run"));
+	N3_VERIFY_UI_COMPONENT(m_pBtn_Act_Stop, GetChildByID("btn_stop"));
+	N3_VERIFY_UI_COMPONENT(m_pBtn_Act_Attack, GetChildByID("btn_attack"));
+	N3_VERIFY_UI_COMPONENT(m_pBtn_Act_StandUp, GetChildByID("btn_stand"));
+	N3_VERIFY_UI_COMPONENT(m_pBtn_Act_SitDown, GetChildByID("btn_sit"));
+
+	// 일어서기 버튼은 미리 죽여놓는다..
+	if(m_pBtn_Act_StandUp)
```
twostars:
```cpp
if (m_pBtn_Act_StandUp != nullptr)
```

Client/WarFare/UICmd.cpp
```diff
@@ -262,7 +262,7 @@ bool CUICmd::OnKeyPress(int iKey)
 			//열려있는 다른 유아이를 닫아준다.
 			CGameProcedure::s_pUIMgr->ReFocusUI();//this_ui
 			CN3UIBase* pFocus = CGameProcedure::s_pUIMgr->GetFocusedUI();
-			if(pFocus && pFocus != this) pFocus->OnKeyPress(iKey);
+			if(pFocus != nullptr && pFocus != this) pFocus->OnKeyPress(iKey);
```
twostars:
```cpp
if (pFocus != nullptr && pFocus != this)
    pFocus->OnKeyPress(iKey);
```

Client/WarFare/UICmd.cpp
koraykarakus: i need more important questions about that ? why buttons used as UIBase and which type we supposed to use for bottons overall in all project. 

choise 1 : use as Base so type conversion is not needed.
choise 2 : use as Button - include button & convert type.

none of these PRs focusing spacing, it can be handled seperatelly later. Objective is to use nullptr & N3_VERIFY macro overall project so that new PRs submitted in the same way. I saw people still use old way to load components and use NULL instead of nullptr and that is caused by variety of the code on UI.

twostars: They were used like this because they had no need to access any CN3UIButton-specific logic.
They're still "loaded" the same. It's just how you're accessing them. There's some cases where they're not even the types they suggest (CUISkillTreeDlg is a big example of this).

> none of these PRs focusing spacing, it can be handled seperatelly later.

If you're updating it, it should just be done at the same time. No reason to go back over it again in the future.

Client/WarFare/UICmd.cpp
```diff
@@ -56,31 +56,31 @@ CUICmd::~CUICmd()
...
-	m_pBtn_Party_Disband =	GetChildByID("btn_disband");	//__ASSERT(m_pBtn_Party_Disband, "NULL UI Component!!");
-	m_pBtn_Skill =			GetChildByID("btn_skill");		//__ASSERT(m_pBtn_Skill, "NULL UI Component!!");
-	m_pBtn_Exit	=			GetChildByID("btn_Exit");		//__ASSERT(m_pBtn_Exit, "NULL UI Component!!");
-
-	m_pBtn_Quest =		GetChildByID("btn_quest");			//__ASSERT(m_pBtn_Quest, "NULL UI Component!!");
-	m_pBtn_Belong =		GetChildByID("btn_knight");			//__ASSERT(m_pBtn_Belong, "NULL UI Component!!");
+	N3_VERIFY_UI_COMPONENT(m_pBtn_Act, GetChildByID("btn_control"));
+	N3_VERIFY_UI_COMPONENT(m_pBtn_Act_Walk, GetChildByID("btn_walk"));
+	N3_VERIFY_UI_COMPONENT(m_pBtn_Act_Run, GetChildByID("btn_run"));
+	N3_VERIFY_UI_COMPONENT(m_pBtn_Act_Stop, GetChildByID("btn_stop"));
```
twostars: This button doesn't exist in the UI. Its assertion was disabled before, but now it'll trigger since it doesn't exist.

Client/WarFare/UICmd.cpp
```diff
@@ -56,31 +56,31 @@ CUICmd::~CUICmd()
...
+		m_pBtn_Act_StandUp->SetVisible(false); 
+	
+	N3_VERIFY_UI_COMPONENT(m_pBtn_Character, GetChildByID("btn_character"));
+	N3_VERIFY_UI_COMPONENT(m_pBtn_Inventory, GetChildByID("btn_inventory"));
+	N3_VERIFY_UI_COMPONENT(m_pBtn_CmdList, GetChildByID("btn_option"));
+	N3_VERIFY_UI_COMPONENT(m_pBtn_Camera, GetChildByID("btn_camera"));
+	N3_VERIFY_UI_COMPONENT(m_pBtn_Party_Invite, GetChildByID("btn_invite"));
+	N3_VERIFY_UI_COMPONENT(m_pBtn_Party_Disband, GetChildByID("btn_disband"));
+	N3_VERIFY_UI_COMPONENT(m_pBtn_Skill, GetChildByID("btn_skill"));
+	N3_VERIFY_UI_COMPONENT(m_pBtn_Exit, GetChildByID("btn_Exit"));
```
twostars: This is just a note while I'm checking these in the UI, but this should just be btn_exit (it's how it's named in the UI itself). The comparison is case-insensitive so it doesn't matter, but if we were to speed this up in future with a case-sensitive lookup, this would need to be changed.

Might as well fix that one now.

Client/WarFare/UICmd.cpp
```diff
@@ -56,31 +56,31 @@ CUICmd::~CUICmd()
...
+	N3_VERIFY_UI_COMPONENT(m_pBtn_Quest, GetChildByID("btn_quest"));
```
twostars: This button doesn't exist in the UI. Its assertion was disabled before, but now it'll trigger since it doesn't exist.

Client/WarFare/UICmd.cpp
```diff
@@ -56,31 +56,31 @@ CUICmd::~CUICmd()
...
+	N3_VERIFY_UI_COMPONENT(m_pBtn_Belong, GetChildByID("btn_knight"));
```
twostars: This button doesn't exist in the UI. Its assertion was disabled before, but now it'll trigger since it doesn't exist.

Client/WarFare/UICmd.cpp
twostars: To answer your question though, typically we use the explicit type so it's clear.
Any other behaviour is just old outdated mgame behaviour. But you just have to make sure that it is in fact the suggested type before changing it, just in case.

Note that I didn't go through and mass-replace these because the implementation is still iffy.

Ideally we should double-check the type is valid, rather than just blindly casting and hoping for the best, which the macro doesn't really allow for.

twostars: I've just merged in template specializations for this. So they can now just use GetChildByID\<type\> for the specific types (don't bother supplying a type if you just want CN3UIBase).

These are more strictly type checked and the change fits with the existing macros.

twostars (approved):

PR #445, opened 2025-08-17
UICmdEdit to N3_VERIFY_UI_COMPONENT & spacing

koraykarakus:
## Pull request type

uses __ASSERT macro 

## What is the new behaviour?

N3_VERIFY macro is used

UI elements should be loaded same in all sections of client.

## Checklist

### DIFF
--- a/Client/WarFare/UICmdEdit.cpp
+++ b/Client/WarFare/UICmdEdit.cpp
@@ -33,12 +33,14 @@ CUICmdEdit::~CUICmdEdit()
 
 bool CUICmdEdit::Load(HANDLE hFile)
 {
-	if (CN3UIBase::Load(hFile) == false) return false;
+	if (CN3UIBase::Load(hFile) == false) 
+		return false;
+	
+	N3_VERIFY_UI_COMPONENT(m_pText_Title, (CN3UIString*) GetChildByID("Text_cmd"));
+	N3_VERIFY_UI_COMPONENT(m_pBtn_Ok,	  (CN3UIButton*) GetChildByID("btn_ok"));
+	N3_VERIFY_UI_COMPONENT(m_pBtn_Cancel, (CN3UIButton*) GetChildByID("btn_cancel"));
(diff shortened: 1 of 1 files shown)


Client/WarFare/UICmdEdit.cpp
```diff
@@ -33,12 +33,14 @@ CUICmdEdit::~CUICmdEdit()
 
 bool CUICmdEdit::Load(HANDLE hFile)
 {
-	if (CN3UIBase::Load(hFile) == false) return false;
+	if (CN3UIBase::Load(hFile) == false)
```
twostars:
```cpp
if (!CN3UIBase::Load(hFile))
```

Client/WarFare/UICmdEdit.cpp
```diff
@@ -33,12 +33,14 @@ CUICmdEdit::~CUICmdEdit()
 
 bool CUICmdEdit::Load(HANDLE hFile)
 {
-	if (CN3UIBase::Load(hFile) == false) return false;
+	if (CN3UIBase::Load(hFile) == false) 
+		return false;
+	
+	N3_VERIFY_UI_COMPONENT(m_pText_Title, (CN3UIString*) GetChildByID("Text_cmd"));
```
twostars: Please use tabs for indentation. Additionally, for type safety, this should be `GetChildByID<CN3UIString>("Text_cmd")` now.

Client/WarFare/UICmdEdit.cpp
```diff
@@ -77,7 +79,8 @@ void CUICmdEdit::Open(std::string msg)
 
 void CUICmdEdit::SetVisible(bool bVisible)
 {
-	if (bVisible == this->IsVisible()) return;
+	if (bVisible == this->IsVisible())
```
twostars: `this->` can go.

Client/WarFare/UICmdEdit.cpp
koraykarakus: do we have templete overload now ? if yes nice , if no C style cast should be enough because we are certain about the type of text_cmd.

twostars: As of #441 we do.

PR #450, opened 2025-08-19
CUIDroppedItemDlg: Reformatter pass and cleanup

koraykarakus:
## Pull request type

NULL to nullptr, log in debug only, spacing, N3_VERIFY_UI_COMPONENT

## Checklist


twostars:
> log in debug only

Let's not do that. These are logged officially as well in release builds.

twostars:
Also, you seem to be doing a lot of this manually. There's no need to do that. CTRL+K+F reformats the selected text. CTRL+K+F reformats the entire document. It doesn't catch everything, but it fixes up most of the spacing issues (it just loses some special indentation along the way that you'd have to revert).

PR #453, opened 2025-08-19
CUser::GoldGain(): Fix overflow

koraykarakus:
## Pull request type

- [x] Bugfix

gold gain overflows

## What is the new behaviour?

sum as int64_t , overflow check with INT_MAX, check negative gold

This bug can be result in negative gold.

## Checklist

### DIFF
--- a/Server/Ebenezer/User.cpp
+++ b/Server/Ebenezer/User.cpp
@@ -9291,7 +9291,6 @@ void CUser::GoldChange(short tid, int gold)
 		}
 	}
 
-	// First the source...
 	SetByte(send_buff, WIZ_GOLD_CHANGE, send_index);
 	SetByte(send_buff, s_type, send_index);
 	SetDWORD(send_buff, s_temp_gold, send_index);
@@ -11632,7 +11631,6 @@ void CUser::GoldGain(int gold)
 	// set user gold as iTotalGold
 	m_pUserData->m_iGold = static_cast<int>(iTotalGold);
 
-	// First the source...
 	SetByte(send_buff, WIZ_GOLD_CHANGE, send_index);
 	SetByte(send_buff, GOLD_CHANGE_GAIN, send_index);
 	SetDWORD(send_buff, gold, send_index);
--- a/Server/Ebenezer/User.cpp
+++ b/Server/Ebenezer/User.cpp
@@ -11626,16 +11626,15 @@ void CUser::GoldGain(int gold)
 	
 	iTotalGold = static_cast<int64_t>(m_pUserData->m_iGold) + static_cast<int64_t>(gold);
 
-	if (iTotalGold > INT_MAX)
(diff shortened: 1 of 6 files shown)


twostars:
One suggestion: it specifically caps it at 2.1bil, not INT_MAX.

Server/Ebenezer/User.cpp
```diff
@@ -11616,11 +11616,26 @@ void CUser::GoldGain(int gold)
...
+	if (iTotalGold > INT_MAX)
+		iTotalGold = INT_MAX;
+
+	// set user gold as iTotalGold
+	m_pUserData->m_iGold = static_cast<int>(iTotalGold);
+
+	// First the source...
+	SetByte(send_buff, WIZ_GOLD_CHANGE, send_index);	
+	// 1 -> Get gold    2 -> Lose gold
+	SetByte(send_buff, 1, send_index);
```
twostars: This should be an enum for the gold change type.

Server/Ebenezer/User.cpp
```diff
@@ -11616,11 +11616,26 @@ void CUser::GoldGain(int gold)
...
+		m_pUserData->m_iGold = 0;
 
-	SetByte(send_buff, WIZ_GOLD_CHANGE, send_index);	// First the source...
-	SetByte(send_buff, 1, send_index);	// 1 -> Get gold    2 -> Lose gold
+	if (gold < 0)
+		gold = 0;
+	
+	iTotalGold = static_cast<int64_t>(m_pUserData->m_iGold) + static_cast<int64_t>(gold);
+
+	if (iTotalGold > INT_MAX)
```
twostars: This should cap at 2.1bil, not INT_MAX.

twostars (changes requested):

Server/shared/globals.h
```diff
@@ -310,5 +310,11 @@ enum e_AttackResult
 	MAGIC_ATTACK_TARGET_DEAD	= 4
 };
 
+enum e_NoahChangeType
```
twostars: Why is it NoahChangeType and then uses "GOLD_XYZ"?
Either make it e_NoahChangeType and have them explicitly named `NOAH_CHANGE_GAIN`, `NOAH_CHANGE_LOSE`, or switch it to e_GoldChangeType (which mimics the opcode's name) and have them named `GOLD_CHANGE_GAIN`, `GOLD_CHANGE_LOSE`).

(though 'lose' seems a bit odd here, 'loss' is more natural but it *does* mimic the official method name I guess)

twostars (changes requested):

Server/shared/globals.h
koraykarakus: forgot to change it, I was about to go with noah but so many things using gold, but it is noah in reality. Keep it as gold for now.

twostars (approved):

PR #455, opened 2025-08-20
GoldLose [negative check] [return type]

koraykarakus:
… bool from BOOL

## Pull request type

- [x] Bugfix

Does not check for negative values of both "gold" and "usergold". This can cause overflow when gold is send as negavite ur user having a negative gold.

## What is the new behaviour?

checks for negative gold and set as zero on the encounter. Return type is also changed from BOOL to bool.

to avoid overflow and wrapping around.

## Checklist

### DIFF
--- a/Server/Ebenezer/User.cpp
+++ b/Server/Ebenezer/User.cpp
@@ -11638,14 +11638,20 @@ void CUser::GoldGain(int gold)
 	Send(send_buff, send_index);
 }
 
-BOOL CUser::GoldLose(int gold)
+bool CUser::GoldLose(int gold)
 {
 	int send_index = 0;
 	char send_buff[256] = {};
 
+	if (m_pUserData->m_iGold < 0)
+		m_pUserData->m_iGold = 0;
+
+	if (gold < 0)
+		gold = 0;
+
 	// Insufficient gold!
 	if (m_pUserData->m_iGold < gold)
-		return FALSE;
+		return false;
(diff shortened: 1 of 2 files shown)


Server/Ebenezer/User.cpp
```diff
@@ -11638,14 +11638,20 @@ void CUser::GoldGain(int gold)
 	Send(send_buff, send_index);
 }
 
-BOOL CUser::GoldLose(int gold)
+bool CUser::GoldLose(int gold)
 {
 	int send_index = 0;
 	char send_buff[256] = {};
 
+	if (m_pUserData->m_iGold < 0)
```
twostars, line 11650: I'm really not a fan of this check. I was hesitant about it with GoldGain() as well.
Rather than modify it, just fail to apply. Do the same for GoldGain().

If there's a problem and coins are overflowing, we shouldn't just hide the issue. It should simply fail so the issue can be investigated and fixed, before things get too far along that it's harder to figure out the root cause. There's no reason to reset it.

koraykarakus: yea maybe. I added it because most of my characters which I use to test game had negative coin already.

twostars: You should still fail it in addition to the log though.

koraykarakus: added

PR #459, opened 2025-08-21
object event fail return, packet change to support wrong warp gate me…

koraykarakus:
…ssage

## Pull request type

- [x] Bugfix

Wrong warp gate message is not displayed because packet lack info.

## What is the new behaviour?

required info to display fail added to packet.

Snoxd koserver's behavior imitated.

## Checklist

### DIFF
--- a/Server/Ebenezer/User.cpp
+++ b/Server/Ebenezer/User.cpp
@@ -9790,6 +9790,12 @@ BOOL CUser::WarpListObjectEvent(short objectindex, short nid)
 	if (pEvent == nullptr)
 		return FALSE;
 
+	// If the warp gate belongs to a nation, which isn't us...
+	// or we're in the opposing nation's zone...
+	if ((pEvent->sBelong != 0 && pEvent->sBelong != GetNation())
+		|| (pMap->m_nZoneNumber != GetNation() && pMap->m_nZoneNumber <= ELMORAD))
+		return FALSE;
+
 	if (!GetWarpList(pEvent->sControlNpcID))
 		return FALSE;
 
@@ -9855,6 +9861,7 @@ void CUser::ObjectEvent(char* pBuf)
 
 fail_return:
 	SetByte(send_buff, WIZ_OBJECT_EVENT, send_index);
+	SetByte(send_buff, pEvent == nullptr ? 0 : pEvent->sType, send_index);
 	SetByte(send_buff, 0, send_index);
 	Send(send_buff, send_index);
 }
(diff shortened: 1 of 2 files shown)


twostars:
Don't just blindly copy code from snoxd-koserver. The entire point of the reset was ensuring we think critically about all new additions, making sure they're remotely correct in the first place and aren't just random junk changes thrown in for either newer versions or just in general because whoever added it down-the-track felt like it.

In this case the logic is basically there, but it's still very evident that you just ripped it, comments and all, and tried to restore its accessor.

I tidied it up this time, but please -- only use it as a loose reference. Don't copy code from it.

twostars:
So what you've changed is allowing them to interact with the warp gate, but not actually submitting the warp request (SelectWarpList()).
The warp request itself still allows users to use the other warp gate.

This also needs fixing before this issue can be considered resolved, otherwise it's abusable.

twostars (changes requested):
So what you've changed is allowing them to interact with the warp gate, but not actually submitting the warp request (SelectWarpList()).
The warp request itself still allows users to use the other warp gate.

This also needs fixing before this issue can be considered resolved, otherwise it's abusable.

twostars:
> repeating nullcheck

The thing is, that check wasn't even necessary. Every instance is pre-coupled with the _USER_DATA block. It can never be nullptr.
There's no reason to even check it.

> and accessing via pointer each time to nation info required seems pointless

There's literally no difference here between accessing that, and the map instance; it's doing exactly the same thing, with the exact same cost. I was just referencing the official code here.

I'm not against accessors (I was the one that wrote that code you copied and introduced those to snoxd-koserver in the first place, after all), but it's unnecessary here. It can be done at a later stage, rather than just one random usage here just because you're copying it from elsewhere.

Additionally, your declaration of said accessor needed tidying up anyway:
```cpp
inline BYTE GetNation() const { return (m_pUserData != nullptr) ? m_pUserData->m_bNation : 0; };
```
1. Don't put it on one giant line. As I've told you repeatedly, you can use the formatter on the selected (CTRL+K, F) to format things for you. This will use the .editorconfig settings and tidy it up as it's expected to be.
2. Only method declarations should end in a ';'.  This is an (inline) implementation, not a declaration. So the ';' should go.

Rather than bother to point this out, for one single case that really should just be enforced consistently upon implementation anyway, I just stripped it. It didn't really need to be included when copying the code over.

koraykarakus:
I would care more about BYTE than codestyle. BYTE is platform depended although uint8_t is not. There must be some automated API to handle code formatting while pulling requests. Going over each line and pressing a shortcut to format it should be avoided.

PR #473, opened 2025-08-22
N3TexViewer: Support GTT files

UTengine:
In N3TexViewer, this detects .GTT files and loads all containing textures. Each can be selected through the use of the left/up (prev) and right/down (next) keys.

When converting etc, all will be converted as they're generally assumed to be in the same format and dimensions.

While not really supported by the UI, we also allow for the file list to include the .GTTs -- and shift it out of the current working directory, to the directory of the file we've selected.
Consequently, we remove all references to the current working directory and replace it with the loaded directory.

In order to detect that it loads, we tweak CN3Texture::SkipFileHandle() (and CN3Texture::Load() for consistency), and CN3BaseFileAccess::Load() to detect if we're actually reading anything.
In this way it can fail when it hits EOF.

Additionally, we fix a bug with off-screen surfaces not being created (e.g. for saving to bitmap) as they're not supported in the `D3DPOOL_MANAGED` pool.

Finally, we also fix a cast bug with the alpha texture, causing it to not be generated correctly.

### DIFF
--- a/src/Tools/N3TexViewer/N3TexViewerView.cpp
+++ b/src/Tools/N3TexViewer/N3TexViewerView.cpp
@@ -118,7 +118,7 @@ BOOL CN3TexViewerView::PreTranslateMessage(MSG* pMsg)
 					GetDocument()->SelectNextTexture();
 					return TRUE;
 				}
-				return TRUE;
+				break;
 
 			default:
 				break;


twostars:
I feel like it's a bit misleading for it to be able to load the file, then only load the first of the 8 textures contained within. Likewise when it goes to save.

UTengine:
Was thinking about it myself. But the n3tool def didnt even kick in and since I had needed it thouvht I would just do it like this.

PR #479, opened 2025-09-05
Fix “buffer is too small error” in DBAGENT::GetAllCharID

Mervandeli:
The test  DB  which shared in the README  causes "buffer  is too small error" on the character selection screen.

- [x] Bugfix
- [ ] Feature
- [ ] Build related changes

It throws "buffer too small error" on the character selection screen and characters don’t appear.

Ekran görüntüsü 2025-09-04 013432

## What is the new behaviour?
The characters are visible and selectable.

Added a length check and enforced the max limit.

## Checklist


twostars:
What are your character names set to in ACCOUNT_CHAR? What is the contents of the character names here in the string they're being copied from?

Issue #260, opened 2025-06-29
Ebenezer: Update EVT support


twostars:
Since we're going back to EVTs, the server will need to support at least what's needed for the existing 1.298 EVTs.

For LOGIC_ELSE (A) opcodes the full list as of 1.298 is (which includes opcodes which are implemented already):
 - [ ] CHECK_UNDER_WEIGHT
 - [ ] CHECK_OVER_WEIGHT
 - [ ] CHECK_SKILL_POINT
 - [ ] CHECK_EXIST_ITEM
 - [ ] CHECK_NOEXIST_ITEM
 - [ ] CHECK_CLASS
 - [ ] CHECK_WEIGHT
 - [ ] CHECK_EDITBOX
 - [ ] RAND
 - [ ] CHECK_LV
 - [ ] HOWMUCH_ITEM
 - [ ] NOEXIST_COM_EVENT
 - [ ] EXIST_COM_EVENT
 - [ ] CHECK_NOAH
 - [ ] CHECK_NATION
 - [ ] CHECK_PPCARD_SERIAL
 - [ ] CHECK_PPCARD_TYPE
 - [ ] CHECK_EXIST_ITEM_AND
 - [ ] CHECK_EXIST_ITEM_OR
 - [ ] CHECK_NOEXIST_ITEM_AND
 - [ ] CHECK_NOEXIST_ITEM_OR
 - [ ] CHECK_NOEXIST_EVENT
 - [ ] CHECK_EXIST_EVENT
 - [ ] CHECK_PROMOTION_ELIGIBLE
 - [ ] CHECK_EXCHANGE
 - [ ] CHECK_ITEMCHANGE_NUM
 - [ ] CHECK_NOCLASS
 - [ ] CHECK_LOYALTY
 - [ ] CHECK_CHIEF
 - [ ] CHECK_NO_CHIEF
 - [ ] CHECK_CLAN_GRADE
 - [ ] CHECK_KNIGHT
 - [ ] CHECK_DICE
 - [ ] CHECK_CLAN
 - [ ] CHECK_NO_CLAN
 - [ ] CHECK_MANNER
 - [ ] CHECK_MONSTER_CHALLENGE_TIME
 - [ ] CHECK_MONSTER_CHALLENGE_USERCOUNT
 - [ ] CHECK_CASTLE
 - [ ] CHECK_NO_CASTLE
 - [ ] CHECK_SKILL_TOTAL
 - [ ] CHECK_STAT_TOTAL
[...]

PR #498, opened 2025-09-08
Implement CHECK_NOEXIST_ITEM

stefannikolei:
Implemented CHECK_NOEXISTS_ITEM

Implemented parsing of this:
EVENT 8010
A CHECK_NOEXIST_ITEM 910014000 5
E SAY -1 -1 10310 -1 -1 -1 -1 -1 -1 -1 ; Àç·áºÎÁ·
E RETURN
END

We got 2 parameters. the item no and the item count.

Also implemented the logic in User.cpp (just negated the check_exists_item)

#260


twostars:
The only reason the enum values are explicitly set like this is to match the official binary (for reversing purposes).
I would just shift LOGIC_NOEXIST_ITEM to the end so there's only a minor change there; I'll add the rest of them (and sync its value) in a separate PR.

PR #512, opened 2025-09-11
Info font should be bold

### LINKED ISSUE #501: GM NPC overhead text should be bold
As per title. It's bold in the official client, but not here.

twostars:
The info font is just intended to always be bold.

Resolves #501

### DIFF
--- a/Client/WarFare/PlayerBase.cpp
+++ b/Client/WarFare/PlayerBase.cpp
@@ -329,11 +329,11 @@ void CPlayerBase::InfoStringSet(const std::string& szInfo, D3DCOLOR crFont)
 	{
 		std::string szFontInfo = fmt::format_text_resource(IDS_FONT_INFO);
 
-		m_pInfoFont = new CDFont(szFontInfo, 12);
-		m_pInfoFont->InitDeviceObjects( s_lpD3DDev );
+		m_pInfoFont = new CDFont(szFontInfo, 12, D3DFONT_BOLD);
+		m_pInfoFont->InitDeviceObjects(s_lpD3DDev);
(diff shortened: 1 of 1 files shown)


Client/WarFare/PlayerBase.cpp
```diff
@@ -329,11 +329,11 @@ void CPlayerBase::InfoStringSet(const std::string& szInfo, D3DCOLOR crFont)
...
 		std::string szFontInfo = fmt::format_text_resource(IDS_FONT_INFO);
 
-		m_pInfoFont = new CDFont(szFontInfo, 12);
-		m_pInfoFont->InitDeviceObjects( s_lpD3DDev );
+		m_pInfoFont = new CDFont(szFontInfo, 12, D3DFONT_BOLD);
+		m_pInfoFont->InitDeviceObjects(s_lpD3DDev);
 		m_pInfoFont->RestoreDeviceObjects();
 	}
-	m_pInfoFont->SetText(szInfo.c_str(), D3DFONT_BOLD); // 폰트에 텍스트 지정.
+	m_pInfoFont->SetText(szInfo, D3DFONT_BOLD); // 폰트에 텍스트 지정.
```
stefannikolei, line 336: I also looked at this. But I thought oh ok here we got bold. But I have never looked at the font definition.. 

When we don't pass bold here we get a non bold text then? 

The only problem was that the font was not declared as bold?

twostars: The font flag is useless here, I just didn't bother to strip it. The single flag that can be used here is D3DFONT_FILTERED, that's it (and technically D3DFONT_CENTERED, but neither we nor they ever use its 3D text implementation).

PR #586, opened 2025-09-11
Add command handler, implement +zonechange

### LINKED ISSUE #553: Server: Implement GM command +zonechange

twostars:
Implement a command handler generally mimicking official behaviour, except we pre-parse all args and use a compile-time pre-hashed array for ease-of-use & performance.

Implement +zonechange as the first such case.

Resolves #553

### DIFF
--- a/Server/Ebenezer/EbenezerDlg.cpp
+++ b/Server/Ebenezer/EbenezerDlg.cpp
@@ -2255,7 +2255,7 @@ BOOL CEbenezerDlg::PreTranslateMessage(MSG* pMsg)
 			if (chatstr[0] == '+')
 			{
 				OperationMessage opMessage(this, nullptr);
-				opMessage.ParseGM(chatstr);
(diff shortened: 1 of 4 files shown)


Server/Ebenezer/EbenezerDlg_CommandHandler.cpp
```diff
@@ -0,0 +1,102 @@
...
+			spdlog::warn(
+				"EbenezerDlg::OperationMessage: parsed argument out of range from server [command={}, exception={}]",
+				command, ex.what());
+		}
+	}
+}
+
+// +zonechange: {int: zoneId} [float: x] [float: z]
+// NOTE: Coordinates are unofficial.
+void CEbenezerDlg::OperationMessage_ZoneChange(const char* command, CUser* pSrcUser, const std::vector<std::string>& parts)
```
kenner2: looks like the user pointer can be made const

PR #599, opened 2025-09-22
Implement CHECK_CLAN_GRADE

stefannikolei:
* Added implementation for checking the clan grade

#260

### DIFF
--- a/Server/Ebenezer/User.cpp
+++ b/Server/Ebenezer/User.cpp
@@ -12559,19 +12559,19 @@ BOOL CUser::CheckItemCount(int itemid, short min, short max)
 	return FALSE;
 }
 
-BOOL CUser::CheckClanGrade(short min, short max)
+bool CUser::CheckClanGrade(int min, int max)
 {
 	if (m_pUserData->m_bKnights == 0)
-		return FALSE;
+		return false;
 
 	CKnights* pKnights = m_pMain->m_KnightsMap.GetData(m_pUserData->m_bKnights);
 	if (pKnights == nullptr)
-		return FALSE;
+		return false;
 
 	if (pKnights->m_byGrade < min || pKnights->m_byGrade > max)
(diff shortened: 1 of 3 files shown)


Server/Ebenezer/User.cpp
```diff
@@ -12554,6 +12559,21 @@ BOOL CUser::CheckItemCount(int itemid, short min, short max)
 	return FALSE;
 }
 
+BOOL CUser::CheckClanGrade(short min, short max)
```
twostars: 2 things:
1. We can just use the native C++ type 'bool' here, rather than the Windows-specific WINAPI type BOOL (and then FALSE -> false, TRUE -> true respectively).
2. The type casts here are a bit interesting: you're having it demoted from int -> short, and then below, comparison below is promoting from BYTE to short.
We can go one of 2 ways with this, but only one would be technically official; either we:
  - demote the checked value to BYTE (or rather, uint8_t is the appropriate C++ standard type), causing it to not behave the same with certain inputs, or we
  - handle it like official and promote min/max to int (or well, just keep min/max int so the data is passed through as-is), and let it upcast m_byGrade to int for comparison, which would ensure official behaviour of all values.

I'm personally more in favour of option 2 just for consistency with official check behaviour, even though for real usage it doesn't really matter too much.

Either way, the cast should be done explicitly - it will be producing a warning as-is, which we'll want to be avoiding.

stefannikolei: Should't the native c++ bool values be used throughout the whole project? 

I just used it because GetItemCount used them. That's why I went with those. 

The same applies to min and max. Copied it from the method above.

Will implement the second option you have outlined.

twostars: They should, yes. min and max in the above, however, refer to the item count, which is short already so there's no awkward casting there.

PR #600, opened 2025-09-22
Implement LOGIC_CHECK_EMPTY_SLOT

stefannikolei:
#260

### DIFF
--- a/Server/Ebenezer/User.cpp
+++ b/Server/Ebenezer/User.cpp
@@ -8678,16 +8678,16 @@ void CUser::InitType4()
 
 int CUser::GetNumberOfEmptySlots() const
 {
-	int emptySlotsCount = 0;
+	int emptySlotCount = 0;
 
 	for (int i = SLOT_MAX; i < SLOT_MAX + HAVE_MAX; i++)
 	{
-		const _ITEM_DATA& pItem = m_pUserData->m_sItemArray[i];
-
-		if (pItem.nNum == 0)
-			emptySlotsCount++;
+		const _ITEM_DATA& item = m_pUserData->m_sItemArray[i];
+		if (item.nNum == 0)
+			++emptySlotCount;
 	}
-	return emptySlotsCount;
+
+	return emptySlotCount;
 }
 
 // item 먹을때 비어잇는 슬롯을 찾아야되...


Server/Ebenezer/User.cpp
```diff
@@ -8677,6 +8677,20 @@ void CUser::InitType4()
 	m_bType4Flag = FALSE;
 }
 
+int CUser::GetNumberOfEmptySlots()
+{
+	int emptySlotsCount = 0;
+
+	for(int i = SLOT_MAX; i < SLOT_MAX + HAVE_MAX; i++)
+	{
+		_ITEM_DATA pItem = m_pUserData->m_sItemArray[i];
```
twostars: This creates a copy; use a const reference here instead

Server/Ebenezer/User.cpp
```diff
@@ -8677,6 +8677,20 @@ void CUser::InitType4()
 	m_bType4Flag = FALSE;
 }
 
+int CUser::GetNumberOfEmptySlots()
+{
+	int emptySlotsCount = 0;
+
+	for(int i = SLOT_MAX; i < SLOT_MAX + HAVE_MAX; i++)
```
twostars: Select the method and run a quick formatter pass (press CTRL+K, then press F) over it.

Server/Ebenezer/User.cpp
```diff
@@ -11515,6 +11529,11 @@ BOOL CUser::CheckEventLogic(EVENT_DATA* pEventData)
 					bExact = TRUE;
 				break;
 
+			case LOGIC_CHECK_EMPTY_SLOT:
+				if(GetNumberOfEmptySlots() >= pLE->m_LogicElseInt[0])
```
twostars: Select this and run a quick formatter pass (press CTRL+K, then press F) over it.

Server/Ebenezer/User.h
```diff
@@ -282,6 +282,7 @@ class CUser : public CIOCPSocket2
 	void Home();
 	void ReportBug(char* pBuf);
 	int GetEmptySlot(int itemid, int bCountable);
+	int GetNumberOfEmptySlots();
```
twostars: Mark this as const (at the end), so it knows that we're not actually modifying class members here

Server/Ebenezer/User.cpp
stefannikolei: How can I do that? Still learning things for c++🫣

twostars: It should just be:
```cpp
const _ITEM_DATA& itemData = m_pUserData->m_sItemArray[i];
```

twostars (approved):

PR #605, opened 2025-09-22
Implement CHECK_EXIST_EVENT

stefannikolei:
* Implemented the logic for CHECK_EXIST_EVENT
* Moved it out of #if0 in LOGIC_ELSE.cpp (also moved forgotten CHECK_EMPTY_SLOT)

#260

### DIFF
--- a/Server/Ebenezer/User.cpp
+++ b/Server/Ebenezer/User.cpp
@@ -8690,6 +8690,16 @@ int CUser::GetNumberOfEmptySlots() const
 	return emptySlotCount;
 }
 
+bool CUser::CheckExistEvent(int questId, int questState) const
+{
+	for (const auto& quest : m_pUserData->m_quests)
+	{
+		if (quest.sQuestID == questId && quest.byQuestState == questState)
+			return true;
+	}
+	return false;
+}
+
 // item 먹을때 비어잇는 슬롯을 찾아야되...
 int CUser::GetEmptySlot(int itemid, int bCountable) const
 {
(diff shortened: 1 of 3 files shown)


Server/Ebenezer/User.cpp
```diff
@@ -8690,6 +8690,16 @@ int CUser::GetNumberOfEmptySlots() const
 	return emptySlotCount;
 }
 
+bool CUser::CheckExistEvent(int questId, int questState) const
+{
+	for (const auto& quest : m_pUserData->m_quests)
+	{
+		if (quest.sQuestID == questId && quest.byQuestState == questState)
```
twostars: The quest state check here should decide the result; if it encounters the quest ID and it doesn't match, it shouldn't keep processing. It should just return immediately either way.

Server/Ebenezer/User.cpp
```diff
@@ -8690,6 +8690,16 @@ int CUser::GetNumberOfEmptySlots() const
...
+			return true;
+	}
+	return false;
```
twostars: It should return true at this point if we're requesting `QUEST_STATE_NOT_STARTED`, as there's no quest state for the associated quest yet.

stefannikolei: QUEST_STATE_NOT_STARTED  = 0?

could not find any definition

stefannikolei: so the method should be like this then?
```c++
for (const auto& quest : m_pUserData->m_quests)
{
	if (quest.sQuestID != questId)
		continue;

	if (quest.byQuestState == questState)
		return true;
	break;
}
return questState == 0 ? true : false;
```

twostars: Sorry, too used to too many different codebases. There probably isn't yet an enum defined for the quest states, but it should look something like:
```cpp
enum e_QuestState : uint8_t
{
    QUEST_STATE_NOT_STARTED = 0,
    QUEST_STATE_IN_PROGRESS = 1,
    QUEST_STATE_COMPLETE = 2
};
```

I would throw it in `shared/globals.h` as the client may have need for it.

PR #607, opened 2025-09-23
Implement CHECK_KNIGHT

stefannikolei:
#260

### DIFF
--- a/Server/Ebenezer/User.cpp
+++ b/Server/Ebenezer/User.cpp
@@ -11571,6 +11571,7 @@ bool CUser::CheckEventLogic(const EVENT_DATA* pEventData)
 			case LOGIC_CHECK_KNIGHT:
 				if (CheckKnight())
 					bExact = true;
+				break;
 
 			default:
 				return false;
@@ -12631,17 +12632,11 @@ bool CUser::CheckClanGrade(int min, int max) const
 
 bool CUser::CheckKnight() const
 {
-	if (m_pUserData->m_bKnights == 0)
-		return false;
-
-	CKnights* pKnights = m_pMain->m_KnightsMap.GetData(m_pUserData->m_bKnights);
+	CKnights* pKnights = m_pMain->GetKnightsPtr(m_pUserData->m_bKnights);
 	if (pKnights == nullptr)
 		return false;
 
-	if (pKnights->m_byfFlag == KNIGHTS_TYPE)
-		return true;
-
-	return false;
+	return (pKnights->m_byFlag == KNIGHTS_TYPE);
 }
 
 void CUser::SaveComEvent(int eventid)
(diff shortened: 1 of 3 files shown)


Server/Ebenezer/User.cpp
```diff
@@ -11568,6 +11568,10 @@ bool CUser::CheckEventLogic(const EVENT_DATA* pEventData)
 					bExact = true;
 				break;
 
+			case LOGIC_CHECK_KNIGHT:
+				if (CheckKnight())
+					bExact = true;
+
```
twostars, line 11575: Missing the break statement here; it will fall through and return false.

Server/Ebenezer/User.cpp
```diff
@@ -12625,6 +12629,21 @@ bool CUser::CheckClanGrade(int min, int max) const
 	return true;
 }
 
+bool CUser::CheckKnight() const
+{
+	if (m_pUserData->m_bKnights == 0)
```
twostars: There's a case where this can be set to -1, which would also be considered invalid.
Realistically what's done officially is they move these lookups into `CEbenezerDlg::GetKnightsPtr()` (and it's not actually a map anymore, which is... a choice), but what you could do is mock up GetKnightsPtr() to just handle the map lookup, and preempt with your check here (but <= 0, rather than just == 0) to avoid the unnecessary lookup.

Then your logic would (appear) to match official's, and keep the preempted checks consistent for other map lookups.

Server/Ebenezer/User.cpp
```diff
@@ -12625,6 +12629,21 @@ bool CUser::CheckClanGrade(int min, int max) const
...
+		return false;
+
+	CKnights* pKnights = m_pMain->m_KnightsMap.GetData(m_pUserData->m_bKnights);
+	if (pKnights == nullptr)
+		return false;
+
+	if (pKnights->m_byfFlag == KNIGHTS_TYPE)
```
twostars: That's a typo. That should be m_byFlag.
But I also don't particularly like flipping from early return failures to it succeeding here.

Just ending it with:
```cpp
return (pKnights->m_byFlag == KNIGHTS_TYPE);
```
is probably the nicest way to address that.

Server/Ebenezer/User.cpp
stefannikolei: This should then probably also be changed for CheckClanGrade?

twostars: Yeah, I guess so. I didn't really think to mention it before.

stefannikolei: I don't get it 🙈 I should add a method to ebenezerdlg how can I get the userdata in that class?

twostars: You wouldn't, you'd just add the wrapper, like:
```cpp
CKnights* CEbenezerDlg::GetKnightsPtr(int16_t clanId)
{
    if (clanId <= 0)
        return nullptr;

    return m_KnightsMap.GetData(clanId);
}
```

Then you'd call that as normal and the lookups would be consistent across the board and we'd allow for updating it to behave like it does officially, if need be.

stefannikolei: Yeah ok. That was the solution I was coming up with. But your definition without arguments confused me 😇🤣

PR #610, opened 2025-09-24
Implement CheckPromotionEligible

stefannikolei:
* Add the Check and the Logic for PromotionEligible (ported from the provided example)
* Added SendSay with 10 parameters
* moved e_Class, e_Race, e_Nation from Client to shared
* Used e_Nation throughout the server
#260


Server/Ebenezer/EbenezerDlg.cpp
```diff
@@ -941,10 +941,10 @@ int CEbenezerDlg::GetAIServerPort() const
 {
 	switch (m_nServerNo)
 	{
-		case KARUS:
+		case NATION_KARUS:
```
twostars: These should be in their own enum, rather than using e_Nation. Then "UNIFY_NATION" can be preserved (AFAIK it's  used by the official server, we just don't appear to use it here for some reason) and BATTLE isn't just randomly off by itself.

Server/Ebenezer/User.cpp
```diff
@@ -12069,6 +12073,85 @@ bool CUser::CheckClass(short class1, short class2, short class3, short class4, s
...
+bool CUser::CheckPromotionEligible()
+{
+	CNpc* npc = m_pMain->m_NpcMap.GetData(m_sEventNid);
+	if (npc == nullptr)
+		return false;
+
+	short tempClass = -1;
+
+
+	if (CheckClass(106, 108, 110, 112, -1, -1) || CheckClass(206, 208, 210, 212, -1, -1))
```
twostars: 2 things:
1. Use defaults for these extra args
2. Make use of e_Class here to supply the class names

Server/Ebenezer/User.cpp
```diff
@@ -12069,6 +12073,85 @@ bool CUser::CheckClass(short class1, short class2, short class3, short class4, s
...
+	{
+		if (m_pUserData->m_sClass > 206)
```
twostars: This is very much straight from IDA code; the logic should be tidied up for this to be sensible.
tempClass shouldn't be necessary here at all.

Server/Ebenezer/User.cpp
```diff
@@ -12069,6 +12073,85 @@ bool CUser::CheckClass(short class1, short class2, short class3, short class4, s
...
+		{
+			tempClass = m_pUserData->m_sClass - 208;
+		}
+		else
+		{
+			if (m_pUserData->m_sClass == 206 || m_pUserData->m_sClass == 106)
+			{
+				SendSay(-1, -1, 6006, -1, -1, -1, -1, -1, -1, -1);
```
twostars: These can also just use defaults for its extra args, though be aware that the first few args should be mandatory.
From memory they're the up & OK buttons and then the main text, but it'd have to be double checked.

Server/Ebenezer/User.cpp
```diff
@@ -12069,6 +12073,85 @@ bool CUser::CheckClass(short class1, short class2, short class3, short class4, s
...
+				return false;
+			}
+			tempClass = m_pUserData->m_sClass - 108;
+		}
+
+		if (tempClass == 0)
```
twostars: The above logic translates this to handle for class 208 or class 108.

Server/Ebenezer/User.cpp
```diff
@@ -12069,6 +12073,85 @@ bool CUser::CheckClass(short class1, short class2, short class3, short class4, s
...
+		{
+			SendSay(-1, -1, 7006, -1, -1, -1, -1, -1, -1, -1);
+		}
+		else
+		{
+			if (tempClass - 2 == 2)
```
twostars: If (class - 108) or (class - 208) is 2, then that makes this 110 or 210.

Server/Ebenezer/User.cpp
stefannikolei: Yeah going to clean it up

Server/Ebenezer/EbenezerDlg.cpp
stefannikolei: Ok so e_Nation back to the client? How should I call the new enum and which values should go in there? I probably misunderstood what you said in discord

twostars: No, the other cases are in reference to player nations. This is in reference to the server... mode, I guess.

stefannikolei: So what do you suggest? When I am honest I have no clue 😆

twostars: So on AI side where this is also a thing, you see that it uses all of them:
```cpp
	if (m_byZone == UNIFY_ZONE)
		m_strStatus.Format(_T("Server Zone: UNIFY"));
	else if (m_byZone == KARUS_ZONE)
		m_strStatus.Format(_T("Server Zone: KARUS"));
	else if (m_byZone == ELMORAD_ZONE)
		m_strStatus.Format(_T("Server Zone: ELMORAD"));
	else if (m_byZone == BATTLE_ZONE)
		m_strStatus.Format(_T("Server Zone: BATTLE"));
```

```cpp
/// \brief fetches the listen port associated with m_byZone
/// \see m_byZone
/// \returns the associated listen port or -1 if invalid
int CServerDlg::GetListenPortByZone() const
{
	switch (m_byZone)
	{
		case KARUS_ZONE:
		case UNIFY_ZONE:
			return AI_KARUS_SOCKET_PORT;

		case ELMORAD_ZONE:
			return AI_ELMO_SOCKET_PORT;

		case BATTLE_ZONE:
			return AI_BATTLE_SOCKET_PORT;

		default:
			return -1;
	}
}
```

What does this effectively do though.

If we look at what m_byZone in AI does, we see when loading NPCs from the DB, it only handles those in zones managed by it (if 'unify', it handles all, otherwise it handles those in zones attached to this server):
```cpp
	for (model::NpcPos* row : rows)
	{
		bool bMoveNext = true;
		int nPathSerial = 1;
		int nNpcCount = 0;

		do
		{
			int nMonsterNumber = row->NumNpc;
			int nServerNum = GetServerNumber(row->ZoneId);

			if (m_byZone == nServerNum
				|| m_byZone == UNIFY_ZONE)
```

On Ebenezer's side this is -- as in the case this refers to -- m_nServerNo, which interestingly has a few different behaviours depending on what it is.

Conventionally people will just have this in "Karus" mode, which puts us in the typical port 10020 (which is also what "UNIFY_ZONE" uses).

Notably AI defines these independent of the nation itself, though their definition still sucks.

It being named the "server number" is a little confusing because this isn't the same as the "server index". The server index is a unique server ID across all servers in a list.
The "server number" (or AI's server "zone") appears to just dictate the server type, mode, maybe even explicitly 'server zone' (though it's more accurately a *grouping* of zones). I'm not sure what the best terminology to use here would be, to be honest.

But essentially, you'd pull those back into their own enum named distinctly, handled similarly to AI's existing macros (which you'd replace, to be honest), something like -- if we're otherwise preserving their naming -- (again, I don't know what the best terminology to use for this would be):
```cpp
enum e_ServerZone
{
    SERVER_ZONE_UNIFY  = 0, // All zones
    SERVER_ZONE_KARUS  = 1, // Karus-related zones
    SERVER_ZONE_ELMORAD = 2, // El Morad-related zones
    SERVER_ZONE_BATTLE = 3, // War zones (maybe even CZ?)
};
```

But yeah, the point is it's definitely not the nation as used by players/NPCs, so we shouldn't even be remotely considering sharing `e_Nation` for that.

Server/Ebenezer/User.cpp
stefannikolei: CheckClass has no default parameters atm. Should i make the last two as default and set them to -1?

twostars: I would assign defaults to all but the first here, as you would never check it on 0, but you'd want to check at least 1.

Server/Ebenezer/EbenezerDlg.cpp
stefannikolei: ty for the long explanation

stefannikolei: Just found the enum e_ZoneID in globals.h 

It would probably be the best to add Unify and Battle here, or?

stefannikolei: enum e_ZoneID
{
	ZONE_KARUS				= 1,
	ZONE_ELMORAD			= 2,
	ZONE_ESLANT_KARUS		= 11,
	ZONE_ESLANT_ELMORAD		= 12,
	ZONE_MORADON			= 21,
	ZONE_DELOS				= 30,
	ZONE_BIFROST			= 31,
	ZONE_DESPERATION_ABYSS	= 32,
	ZONE_HELL_ABYSS			= 33,
	ZONE_ARENA				= 48,
	ZONE_CAITHAROS_ARENA	= 55,
	ZONE_BATTLE				= 101,
	ZONE_BATTLE2			= 102,
	ZONE_BATTLE3			= 103,
	ZONE_SNOW_BATTLE		= 111,
	ZONE_FRONTIER			= 201
};

twostars: They're not zone IDs either though. They're at best zone *groups*, but not IDs for zones in and of themselves.

stefannikolei: Now while using ne e_ServerZone i get the feeling that it is used in ebenezer as Nation and AIServer as a Zone

twostars: On Ebenezer's side it's the "server number" which happens to explicitly use the same macro as with nation.
On AI it's the "server zone" which uses its own independent macros to denote these types (BATTLE_ZONE, UNIFY_ZONE, KARUS_ZONE, etc).

They're the same thing though. It's all just confusing and misleading.

stefannikolei: 🤯🤣🤣😂

Server/Ebenezer/User.cpp
```diff
@@ -12070,6 +12074,77 @@ bool CUser::CheckClass(short class1, short class2, short class3, short class4, s
...
+		|| CheckClass(CLASS_EL_PROTECTOR, CLASS_EL_ASSASIN, CLASS_EL_ENCHANTER, CLASS_EL_DRUID))
+	{
+		if (m_pUserData->m_sClass == CLASS_EL_PROTECTOR || m_pUserData->m_sClass == CLASS_KA_GUARDIAN)
+		{
+			SendSay(-1, -1, 6006, -1, -1, -1, -1, -1, -1, -1);
+			return false;
+		}
+		else if (m_pUserData->m_sClass == CLASS_EL_ASSASIN || m_pUserData->m_sClass == CLASS_KA_PENETRATOR)
+		{
+			SendSay(-1, -1, 7006, -1, -1, -1, -1, -1, -1, -1);
```
twostars: These should all fall through to return false.
The original code was:
```cpp
    m_sClass = this->m_pUserData->m_sClass;
    if ( m_sClass > 206 )
    {
      v14 = m_sClass - 208;
      v13 = v14 == 0;
    }
    else
    {
      if ( m_sClass == 206 || (v12 = m_sClass - 106) == 0 )
      {
        CUser::SendSay(this, -1, -1, 6006, -1, -1, -1, -1, -1, -1, -1);
        return 0;
      }
      v14 = v12 - 2;
      v13 = v14 == 0;
    }
    if ( v13 )
    {
      CUser::SendSay(this, -1, -1, 7006, -1, -1, -1, -1, -1, -1, -1);
    }
    else
    {
      v15 = v14 - 2;
      if ( v15 )
      {
        if ( v15 == 2 )
          CUser::SendSay(this, -1, -1, 9006, -1, -1, -1, -1, -1, -1, -1);
      }
      else
      {
        CUser::SendSay(this, -1, -1, 8006, -1, -1, -1, -1, -1, -1, -1);
      }
    }
    return 0;
```

Server/Ebenezer/User.cpp
```diff
@@ -12070,6 +12074,77 @@ bool CUser::CheckClass(short class1, short class2, short class3, short class4, s
 	return false;
 }
 
+bool CUser::CheckPromotionEligible()
+{
+	CNpc* npc = m_pMain->m_NpcMap.GetData(m_sEventNid);
+	if (npc == nullptr)
+		return false;
+
+	short tempClass = -1;
```
twostars: This is now unused - when the existing merge conflicts are handled, this should trigger a warning which will fail the build.

Edit: Apparently the warning doesn't trigger here, probably because of the assignment... still, best to remove it.

Server/Ebenezer/User.cpp
```diff
@@ -12127,6 +12202,26 @@ void CUser::SendNpcSay(const EXEC* pExec)
 	Send(send_buff, send_index);
 }
 
+void CUser::SendSay(short m1, short m2, short m3, short m4, short m5, short m6, short m7, short m8, short m9, short m10)
```
twostars: I don't really like these names. I'm assuming it's 'm' for message, but even if that's the case, that's not actually true for the first 2; the first 2 are event IDs (which it will forward back to the server in its response upon clicking), for the 'up' button and the 'ok' button consecutively.

At the very least we should ensure these are named separately so they don't just look like arbitrary message/text IDs.

Server/Ebenezer/User.cpp
```diff
@@ -12070,6 +12074,77 @@ bool CUser::CheckClass(short class1, short class2, short class3, short class4, s
...
+
+	short tempClass = -1;
+
+
+	if (CheckClass(CLASS_KA_GUARDIAN, CLASS_KA_PENETRATOR, CLASS_KA_NECROMANCER, CLASS_KA_DARKPRIEST) 
+		|| CheckClass(CLASS_EL_PROTECTOR, CLASS_EL_ASSASIN, CLASS_EL_ENCHANTER, CLASS_EL_DRUID))
+	{
+		if (m_pUserData->m_sClass == CLASS_EL_PROTECTOR || m_pUserData->m_sClass == CLASS_KA_GUARDIAN)
+		{
+			SendSay(-1, -1, 6006, -1, -1, -1, -1, -1, -1, -1);
```
twostars: Ideally we should take out the trailing -1s to let it use the defaults, and perhaps add a comment for what 6006 says in the client (likewise for the other cases).

Server/Ebenezer/User.cpp
```diff
@@ -12070,6 +12074,77 @@ bool CUser::CheckClass(short class1, short class2, short class3, short class4, s
...
+		}
+		else
+		{
+			return false;
+		}
+	}
+	switch (npc->m_tNpcType)
+	{
+		case NPC_MASTER_WARRIOR:
+			if (m_pUserData->m_sClass != CLASS_KA_BERSERKER && m_pUserData->m_sClass != CLASS_EL_BLADE || m_pUserData->m_bLevel < 60)
```
twostars: Reading this, it's not evident what the order of operation actually is here.
`&&` has higher precedence than `||`, so the order here is:
```cpp
if ((m_pUserData->m_sClass != CLASS_KA_BERSERKER && m_pUserData->m_sClass != CLASS_EL_BLADE)
    || m_pUserData->m_bLevel < 60)
```

I think a constant at least defined in this method for the level here would help things too, since we're just using the magic number 60 throughout.

The resource IDs are also a concern regarding magic numbers but... it's probably OK for now.

twostars (changes requested):

Server/Ebenezer/User.cpp
stefannikolei: Could you tell me the tbl in which the messages are in? Then i can add a comment

twostars: Quest_Talk_us.tbl

stefannikolei: that was much to obvious xD

stefannikolei: I changed the signature of SendSay. I put the EventIds to the end and made them optional. Did you want it this way?

Server/Ebenezer/User.cpp
stefannikolei: Yeah. I made that obvious. Also added a const for 60. 

We could introduce something for the event ids. But this way it is the same as in the evts.

Server/Ebenezer/User.cpp
twostars: I'd rather they matched the original call; that way it reads like it does in the EVT (as well as the official binary). With it being a series of numbers like that, it would be very easy to slip up and not realise they're out of order.

stefannikolei: But you wanted that we can call it with sendsay(6006) right? What signature would you suggest

twostars: I mean the trailing -1s for the optional message IDs can be defaults.
The event IDs should still be attached.

If you really, really wanted, you could do a struct for it, with probably all but the equivalent of 'message1' default initialized to -1, and then pass the struct (by value) instead.
With C++20 you get designated initializers, so you'd be able to call it like:
```cpp
SendSay({
  .EventOk = 123,
  .EventUp = 123,
  .TalkId1 = 123
});
```

Or in this case just:
```cpp
SendSay({ .TalkId1 = 8006 });
```

But honestly it's fine to just keep it matching the official call. That implementation is rarely used as it is.

stefannikolei: omg... i misunderstood trailing.. it could have been so easy..

PR #622, opened 2025-10-15
CUIImageTooltipDlg: Add description & grade

Mervandeli:
## Pull request type

- [ ✓] Bugfix
- [ ] Feature
- [ ] Build related changes

Item description, and item grade/class are not visible.

## What is the new behaviour?
It is displayed correctly like the original client 1298.

## Checklist
- [ ✓] I have performed a self-review of my own code.
- [ ✓] Where applicable, I have checked to make sure that this doesn't introduce incompatible behaviour with the official 1.298 server (e.g. unofficial opcodes or behavioural differences).
- [✓ ] I have checked to make sure that this change does not already exist in the codebase in some fashion (e.g. UI already implemented under a different name).

### DIFF
--- a/Client/WarFare/UIImageTooltipDlg.cpp
+++ b/Client/WarFare/UIImageTooltipDlg.cpp
@@ -1057,25 +1057,33 @@ int	CUIImageTooltipDlg::CalcTooltipStringNumAndWrite(__IconItemSkill* spItem, bo
 	{
 		const std::string& szRemark = spItem->pItemBasic->szRemark;
 	
-		size_t strSize = spItem->pItemBasic->szRemark.size();
-		if (strSize > 0)
+		if (!szRemark.empty())
 		{
-			int splitPos = 0;
-			int spaceCount = 0;
-			for (size_t i = 0; i < strSize; i++)
+			size_t totalWords = 1;
+			for (char c : szRemark)
 			{
-				// Official client: split only on spaces (no '(' split). It only applies the split after 4 words.
-				if (szRemark[i] == ' ')
-					++spaceCount;
-
-				if (szRemark[i] == '(' || (szRemark[i] == ' ' && spaceCount > MAX_WORDS_COUNT))
-				{
-					splitPos = i;
-					break;
-				}
+				if (c == ' ')
+					++totalWords;
 			}
-			if (splitPos > 0)
+
+			if (totalWords >= MIN_WORDS_TO_SPLIT_DESC)
 			{
+				size_t wordsInFirstHalf = (totalWords + 1) / 2;
+				size_t wordsSeen = 1, splitPos = std::string::npos;
+				for (size_t i = 0; i < szRemark.size(); i++)
+				{
+					if (szRemark[i] != ' ')
+						continue;
+
+					if (++wordsSeen > wordsInFirstHalf)
+					{
+						splitPos = i;
+						break;
+					}
+				}
+
+				_ASSERT(splitPos != std::string::npos);
+
 				m_pStr[iIndex]->SetColor(m_CWhite);
 				m_pstdstr[iIndex] = fmt::format("*{}", std::string_view(szRemark.data(), splitPos));
 				m_pStr[iIndex]->SetStyle(UI_STR_TYPE_HALIGN, UISTYLE_STRING_ALIGNCENTER);
@@ -1083,8 +1091,11 @@ int	CUIImageTooltipDlg::CalcTooltipStringNumAndWrite(__IconItemSkill* spItem, bo
 				iIndex++;
 				ERROR_EXCEPTION
 
+				// NOTE: This doesn't perfectly match official's behaviour; they rebuild the lines, always appending a space.
+				// This means that there's a guaranteed space after the word here, before the '*'
+				// To replicate this behaviour, we can manually add a space in the format string, but it's not really necessary.
 				m_pStr[iIndex]->SetColor(m_CWhite);
-				m_pstdstr[iIndex] = fmt::format("{}*", std::string_view(szRemark.data() + splitPos, strSize - splitPos));
+				m_pstdstr[iIndex] = fmt::format("{}*", std::string_view(szRemark.data() + splitPos, szRemark.size() - splitPos));
 				m_pStr[iIndex]->SetStyle(UI_STR_TYPE_HALIGN, UISTYLE_STRING_ALIGNCENTER);
 
 				iIndex++;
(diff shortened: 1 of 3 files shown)


Client/WarFare/UIImageTooltipDlg.cpp
```diff
@@ -1052,6 +1053,79 @@ int	CUIImageTooltipDlg::CalcTooltipStringNumAndWrite(__IconItemSkill* spItem, bo
...
+	}
+	ERROR_EXCEPTION
+
+	if (spItem->pItemExt != nullptr)
+	{
+		e_ItemAttrib eTA = (e_ItemAttrib) (spItem->pItemExt->byMagicOrRare);
+		switch (eTA)
+		{
+			case ITEM_ATTRIB_UNIQUE:
+				m_pStr[iIndex]->SetColor(m_CGold);
```
twostars: This should be m_CGreen as well:
image
image

Client/WarFare/UIImageTooltipDlg.cpp
```diff
@@ -1052,6 +1053,79 @@ int	CUIImageTooltipDlg::CalcTooltipStringNumAndWrite(__IconItemSkill* spItem, bo
 	}
 	ERROR_EXCEPTION
 
+	if (spItem->pItemBasic != nullptr)
+	{
+		std::string szRemark = spItem->pItemBasic->szRemark;
```
twostars: We're not modifying this, so we don't really need to copy it. For aliasing, just use a reference, like:
```cpp
const std::string& szRemark = spItem->pItemBasic->szRemark;
```

Client/WarFare/UIImageTooltipDlg.cpp
```diff
@@ -1052,6 +1053,79 @@ int	CUIImageTooltipDlg::CalcTooltipStringNumAndWrite(__IconItemSkill* spItem, bo
...
+	
+		int strSize = spItem->pItemBasic->szRemark.size();
```
twostars: This should just be `size_t` (`size()` returns `size_t` which is unsigned; `int` is signed).

I would have expected this to throw a warning, but I guess it doesn't purely on assignment; only comparisons.
In the future when we hook up 64-bit builds however, this will most certainly throw a warning (int will be 32-bit, size_t will be 64-bit).

Client/WarFare/UIImageTooltipDlg.cpp
```diff
@@ -1052,6 +1053,79 @@ int	CUIImageTooltipDlg::CalcTooltipStringNumAndWrite(__IconItemSkill* spItem, bo
...
+			}
+			if (splitPos > 0)
+			{
+				m_pStr[iIndex]->SetColor(m_CWhite);
+				m_pstdstr[iIndex] = fmt::format("*{}", szRemark.substr(0, splitPos));
+				m_pStr[iIndex]->SetStyle(UI_STR_TYPE_HALIGN, UISTYLE_STRING_ALIGNCENTER);
+
+				iIndex++;
+
+				m_pStr[iIndex]->SetColor(m_CWhite);
```
twostars, line 1097: iIndex is only "safe" for the first one. ERROR_EXCEPTION catches it (prior to its usage) and skips to `exceptions` if it'll be out of range. The second one *isn't* checked/safe.

As loathe as I am to say it, it will need the ERROR_EXCEPTION macro here to ensure it behaves.

Client/WarFare/UIImageTooltipDlg.cpp
```diff
@@ -1052,6 +1053,79 @@ int	CUIImageTooltipDlg::CalcTooltipStringNumAndWrite(__IconItemSkill* spItem, bo
...
+	{
+		std::string szRemark = spItem->pItemBasic->szRemark;
+	
+		int strSize = spItem->pItemBasic->szRemark.size();
+		if (strSize > 0)
+		{
+			int splitPos = 0;
+			for (int i = 0; i < strSize; i++)
+			{
+				if (szRemark[i] == '(' || (szRemark[i] == ' ' && i > strSize / 2))
```
twostars: Officially this only splits on words dictated by spaces, so it ignores `(`, but so long as it's noted in a comment that it's not actually official behaviour, that's fine.

What it also does, however, is it enforces a minimum word limit of 4 for it to enforce the split, so it isn't splitting on short descriptions.

Client/WarFare/UIImageTooltipDlg.cpp
```diff
@@ -1052,6 +1053,79 @@ int	CUIImageTooltipDlg::CalcTooltipStringNumAndWrite(__IconItemSkill* spItem, bo
...
+				{
+					splitPos = i;
+					break;
+				}
+			}
+			if (splitPos > 0)
+			{
+				m_pStr[iIndex]->SetColor(m_CWhite);
+				m_pstdstr[iIndex] = fmt::format("*{}", szRemark.substr(0, splitPos));
```
twostars: I wouldn't normally mention it, but item tooltips are an unfortunate hotspot for frame drops. This is convenient but also really inefficient, because we're (ignoring the fact that szRemark is copied to begin with) allocating a new string, copying it, then allocating and building a new string from it.

Probably the simplest way to improve this is to change the `substr()` to refer to a `string_view` of the original string.
Something like:
```cpp
m_pstdstr[iIndex] = fmt::format("*{}", std::string_view(szRemark.data(), splitPos));
```
It's slightly more verbose but it's close enough to the original.

This applies to the other case as well.

Client/WarFare/UIImageTooltipDlg.cpp
```diff
@@ -1052,6 +1053,79 @@ int	CUIImageTooltipDlg::CalcTooltipStringNumAndWrite(__IconItemSkill* spItem, bo
...
+			case ITEM_ATTRIB_UNIQUE:
+				m_pStr[iIndex]->SetColor(m_CGold);
+				m_pstdstr[iIndex] = fmt::format_text_resource(IDS_TOOLTIP_UNIQUE);
+				m_pStr[iIndex]->SetStyle(UI_STR_TYPE_HALIGN, UISTYLE_STRING_ALIGNCENTER);
+				iIndex++;
+				break;
+			case ITEM_ATTRIB_UPGRADE:
+				if (spItem->pItemBasic != nullptr)
+				{
+					if (spItem->pItemBasic->byIDK3 == 1)
```
twostars: This is not quite right officially. It's actually this + `pItemExt->bySoulBind`, limited to a total value of 3. The rest of it's fine.

Mervandeli: uint8_t		bySoulBind;					// 30 Soul bind — percentage chance of dropping this item upon death in one-on-one combat; not currently in use.
Is it missing data? I always get 0.

twostars: image

Issue #494, opened 2025-09-07
Server/Cross-platform: Replace MFC (UIs) with console


twostars:
The UI is largely useless and can be swapped to console, so this isn't really an issue. VersionManager is the only one with a UI that does anything, but nobody uses it, so it isn't really much of an issue to just remove. The only UI that's "used" is Ebenezer's, for command input -- which can be replaced via console input.

 - [x] AI server (#635)
 - [x] Ebenezer (#643)
 - [x] Aujard (#642)
 - [x] VersionManager (#638)
 - [x] ItemManager (#641)

PR #635, opened 2025-11-02
Remove MFC/UI:  AIServer

kenner2:
Part of #494 

WIP: I'm putting up into a PR for self-review.

TODO:

- [x] Need some sort of polling/wait mechanism on the serverInstance. Server currently closes when SocketManager::AsyncAccept times out. Polling func should catch this somehow and reset the socket instead of crashing to desktop.
- [x] Need a way to gracefully close the server.  Have to look at catching kill/close signals.
- [x] Remove unused _USER_LOG.t variable and references.
- [x] Confirm its OK to remove DEBUG_NEW.  Find replacement if needed.
- [x] Test `IsPointInRect` as part of Npc PathFinding logic
- [x] Make a pass around classes to clean out redundant MFC comments.
- [x] Rename stdafx.h to pch.h 
- [x] Rename ServerDlg
- [x] Fix communication issues with Ebenezer.
- [x] Restore the string resources (in some capacity) 
- [x] Restore application icon (it must be attached in the .rc file)

### DIFF
--- a/Server/AIServer/AiServerInstance.cpp
+++ b/Server/AIServer/AiServerInstance.cpp
@@ -353,9 +353,8 @@ void AiServerInstance::thread_loop()
 		
 	while (_canTick)
 	{
-		// check for threads that may need to be restarted
-
-		std::this_thread::sleep_for(std::chrono::milliseconds(100));
+		std::unique_lock<std::mutex> lock(_mutex);
+		_cv.wait(lock);
 	}
 }
 


Server/AIServer/AiServer.cpp
```diff
@@ -0,0 +1,31 @@
+﻿// AiServer.cpp : contains the main() function to start the server
+//
+
+#include "stdafx.h"
+#include "ServerDlg.h"
+
+#include <iostream>
```
kenner2: Have to remove this.  Had a cout while trying to get MSBUILD to hit the right entry point.

Server/AIServer/Extern.h
```diff
@@ -16,7 +16,7 @@ struct _PARTY_GROUP
 
 struct _USERLOG
 {
-	CTime t;
+	std::time_t t;
```
kenner2: I should probably be using chrono for this rather than ctime.  Should have no functional impact - t isn't really needed here anymore.  spdlog already handle log timestamps.

Server/AIServer/GameSocket.cpp
```diff
@@ -427,9 +427,8 @@ void CGameSocket::RecvUserInfo(char* pBuf)
 		&& uid < MAX_USER)
 		m_pMain->m_pUser[uid] = pUser;
 
-	_USERLOG* pUserLog = nullptr;
-	pUserLog = new _USERLOG;
-	pUserLog->t = CTime::GetCurrentTime();
+	_USERLOG* pUserLog = new _USERLOG;
+	pUserLog->t = std::time(0);
```
kenner2: Remove this.  userlog.t is set but never used

Server/AIServer/MagicProcess.cpp
```diff
@@ -15,7 +15,6 @@
 #ifdef _DEBUG
 #undef THIS_FILE
 static char THIS_FILE[] = __FILE__;
-#define new DEBUG_NEW
```
kenner2, line 18: @twostars DEBUG_NEW was tied to the MFC libraries and overloaded the new operator to take __FILE__ and __LINE__ as arguments.  Not sure if it's something we want to find a replacement for or not.

Server/AIServer/Npc.cpp
```diff
@@ -3242,8 +3241,9 @@ int CNpc::GetTargetPath(int option)
 	if (targetUser != nullptr)
 	{
 		// Check if user is within search range
-		CRect r(min_x, min_z, max_x + 1, max_z + 1);
-		if (!r.PtInRect(CPoint((int) targetUser->m_curx / TILE_SIZE, (int) targetUser->m_curz / TILE_SIZE)))
+		_RECT r(min_x, min_z, max_x + 1, max_z + 1);
+		if (!IsPointInRect( _POINT(static_cast<int>(targetUser->m_curx / TILE_SIZE),
```
kenner2, line 3242: Need to ensure I didn't break anything here with in-game testing before merging.

Server/AIServer/ServerDlg.cpp
```diff
@@ -1041,85 +877,6 @@ void CServerDlg::StartNpcThreads()
 	m_pZoneEventThread->start();
 }
 
-//	메모리 정리
-BOOL CServerDlg::DestroyWindow()
```
kenner2, line 1045: Logic here was moved to class destructor

Server/AIServer/ServerDlg.h
```diff
@@ -175,13 +143,9 @@ class CServerDlg : public CDialog
 
 // Implementation
 protected:
-	void DefaultInit();
-
-	HICON m_hIcon;
-
 	// Generated message map functions
 	//{{AFX_MSG(CServerDlg)
```
kenner2: Have to make another pass at cleaning the AFX comments out

Server/AIServer/User.cpp
```diff
@@ -24,7 +23,7 @@ static char THIS_FILE[] = __FILE__;
 
 /*
 	 ** Repent AI Server 작업시 참고 사항 **
-	1. Initialize() 수정
+	1. Init() 수정
```
kenner2: Needs revert - false positive on IDE rename func

Server/AIServer/pch.h
kenner2, line 1: Will likely rename this to pch.h before merge

Server/shared-server/GeometricStructs.h
```diff
@@ -0,0 +1,27 @@
+﻿#pragma once
```
kenner2, line 1: @twostars I imagine some of this may exist somewhere else in the codebase.  CN3 equivalents seem to be float based rather than long, so it wouldn't have been a 1:1 replacement.  Let me know if you want me to solve this a different way.

kenner2: I should probably namespace contents of the file.

Server/AIServer/AiServerInstance.cpp
```diff
@@ -211,74 +154,64 @@ BOOL CServerDlg::OnInitDialog()
...
 	
 	//----------------------------------------------------------------------
-	//	Communication Part Initialize ...
+	//	Communication Part Init ...
 	//----------------------------------------------------------------------
-	spdlog::info("ServerDlg::OnInitDialog: initializing sockets");
+	spdlog::info("ServerDlg::Init: initializing sockets");
 	_socketManager.Init(MAX_SOCKET, 0, 1);
-	_socketManager.AllocateServerSockets<CGameSocket>();
+	_socketManager.AllocateServerSockets<CGameSocket>(this);
```
kenner2: Maybe this is what is causing communication issues with Ebenezer somehow.  Each socket needs a reference back to ServerDlg (m_pMain) to do its work.  They were getting their reference via a static MFC function `AfxApp->GetMainWind()`.  I started passing the reference via constructor/init functions as a replacement.

Server/AIServer/AiServerInstance.h
kenner2, line 1: Since the name no longer going to be accurate, will likely rename to `AiServerInstance`; Open to alternate names.

kenner2: (Other server programs will follow the same convention when refactored, i.e. GameServerInstance, LoginServerInstance, AujardServerInstance, ItemManagerServerInstance)

Server/AIServer/MagicProcess.cpp
twostars: Yeah, these entire blocks can go. They're only needed for that.

twostars (approved):

PR #645, opened 2025-11-28
Add Batch Convertion DXT to BMP

Damra:
Update menu item strings translates. kr to en

## Pull request type

   ## N3TexViewer: Batch Convert DXT to BMP Feature
   This PR adds a batch conversion feature to N3TexViewer that allows converting DXT texture files to BMP format.
### Features:
   - Batch conversion dialog with folder selection
   - Recursive DXT file search in Client folder
   - Automatic output directory structure preservation
   - Error logging to conversion_errors.txt
   - Progress bar during conversion
   - All Korean comments translated to English
   
   ### Changes:
   - Added DlgBatchConvertDxtToPng dialog (renamed to BMP)
   - Updated resource IDs from PNG to BMP
   - Implemented recursive file search and directory creation
   - Added comprehensive error handling and logging
- [ x] Bugfix
- [x ] Feature
- [x ] Code style update (formatting, renaming)
- [x ] Refactoring (no functional changes, no API changes)
- [x ] Build related changes
- [x ] Other (please describe):

## Checklist


Client/N3Base/N3Texture.cpp
```diff
@@ -123,7 +123,8 @@ bool CN3Texture::Create(int nWidth, int nHeight, D3DFORMAT Format, BOOL bGenerat
 #endif
 	if(nullptr == m_lpTexture)
 	{
-		__ASSERT(m_lpTexture, "Texture pointer is NULL!");
```
kenner2: An assert should allow you to ignore the error and continue, no?  I don't know if it's a good idea to go without some form of logging here.  We haven't updated the client to use spdlog yet AFAIK, but I think this would make sense as a warn-level log message.

Client/N3Base/N3Texture.cpp
```diff
@@ -896,11 +897,16 @@ bool CN3Texture::SaveToBitmapFile(const std::string& szFN)
 	LPDIRECT3DSURFACE9 lpSurfDest = nullptr;
 	s_lpD3DDev->CreateOffscreenPlainSurface(m_Header.nWidth, m_Header.nHeight, D3DFMT_A8R8G8B8, D3DPOOL_MANAGED, &lpSurfDest, nullptr);
 
-	if(nullptr == lpSurfDest) return false;
+	if(nullptr == lpSurfDest)
+	{
+		lpSurfSrc->Release(); lpSurfSrc = nullptr;
```
kenner2: Please use separate lines for separate instructions

Client/JpegFile/JpegFile.vcxproj
```diff
@@ -28,12 +28,12 @@
   <Import Project="..\..\props\platform_root.props" />
   <PropertyGroup Condition="'$(Configuration)'=='Debug'" Label="Configuration">
     <ConfigurationType>StaticLibrary</ConfigurationType>
-    <PlatformToolset>v143</PlatformToolset>
+    <PlatformToolset>v145</PlatformToolset>
```
kenner2: Was this intentional?  I would assume that we'd want to update the toolset as its own PR with its own pre-release smoke testing, not change it on a whim.

N3ME/N3ME.vcxproj.filters
```diff
@@ -36,123 +36,6 @@
     </Image>
   </ItemGroup>
   <ItemGroup>
-    <ClCompile Include="..\Client\N3Base\BitMapFile.cpp">
```
kenner2: Have to assume this wasn't intentional and that we wouldn't want this removed

N3TexViewer/DlgBatchConvertDxtToPng.cpp
```diff
@@ -0,0 +1,139 @@
+﻿// DlgBatchConvertDxtToPng.cpp : implementation file (Batch Convert DXT to BMP)
+//
+
+#include "stdafx.h"
+#include "N3TexViewer.h"
+#include "DlgBatchConvertDxtToPng.h"
+#include <shlobj.h>
+#include <shlwapi.h>
+#pragma comment(lib, "shlwapi.lib")
```
kenner2: Libraries should be included via the project configuration, not via pragma comments.

N3ME/N3ME.vcxproj.filters
kenner2: @twostars confirmed these changes are OK

N3TexViewer/MainFrm.cpp
```diff
@@ -7,8 +7,19 @@
...
 #include "ProgressBar.h"
 #include <N3Base/BitMapFile.h>
+#include <N3Base/N3Texture.h>
+#include <shlobj.h>
+#include <shlwapi.h>
+#include <io.h>
+#include <direct.h>
+#include <atlconv.h>
+#include <gdiplus.h>
+#pragma comment(lib, "gdiplus.lib")
```
kenner2: Ditto on using project config rather than pragma comments

Client/N3Base/N3Texture.cpp
twostars: Looks like it was removed because the assertion was redundant, since it's enforced after the check and not in conjunction with it. This one's fine to be removed.

Client/JpegFile/JpegFile.vcxproj
twostars: Yeah we don't want to be retargeting everything yet. Install the v143 (VS2022) toolset and it'll build just fine. You can do that with the VS installer (and it'll continue to build within the VS2026 IDE).

twostars:
I added support for automatic selection of the v145 toolset when running under VS2026 (#659). All of those manual tweaks can be stripped, which should address most if not all of the merge conflicts here.

twostars:
> Resolved all issues

Please make sure you push these changes so that we can see and review them.

PR #647, opened 2025-12-03
Add docker as an option for running our latest database

kenner2:
Adds a docker setup that:
1) Sets up an mssql server 2022 instance
2) Has a `kodb-util` service that can be called to perform database operations such as clean imports.  It clones the `kodb-util` tool, builds it, then uses it to import the data from `OpenKO-db` into the mssql instance.

TODO:
- [x] Test on Windows / test using server programs over ODBC.
- [x] Write/test `.cmd` scripts on Windows.
- [X] docker/README.MD


## Pull request type
- [X] Feature

We either pass around a binary backup file, or someone takes the time to learn the `kodb-util` tool to import our data.  This provides an option to those able to install/use docker.

On a painful whim

## Demo
```
$ ./clean_setup.sh
[+] Running 5/5
 ✔ Image knightonline-sqlserver:latest    Removed                                                                                                                                                                                                0.0s 
 ✔ Image knightonline-kodb-util:latest    Removed                                                                                                                                                                                                0.0s 
 ✔ Volume knightonline_sqlserver_scripts  Removed                                                                                                                                                                                                0.0s 
 ✔ Volume knightonline_kodb_util_data     Removed                                                                                                                                                                                                0.0s 
 ✔ Volume knightonline_sqlserver_data     Removed                                                                                                                                                                                                0.0s 
[+] Building 0.3s (15/15) FINISHED                                                                                                                                                                                                                    
 => [internal] load local bake definitions                                                                                                                                                                                                       0.0s
 => => reading from stdin 583B                                                                                                                                                                                                                   0.0s
 => [internal] load build definition from Dockerfile                                                                                                                                                                                             0.0s
 => => transferring dockerfile: 746B                                                                                                                                                                                                             0.0s
 => [internal] load metadata for mcr.microsoft.com/mssql/server:2022-latest                                                                                                                                                                      0.0s
 => [internal] load .dockerignore                                                                                                                                                                                                                0.0s
 => => transferring context: 2B                                                                                                                                                                                                                  0.0s
 => [1/8] FROM mcr.microsoft.com/mssql/server:2022-latest                                                                                                                                                                                        0.0s
 => [internal] load build context                                                                                                                                                                                                                0.0s
 => => transferring context: 71B                                                                                                                                                                                                                 0.0s
 => CACHED [2/8] WORKDIR /var/opt/appscripts                                                                                                                                                                                                     0.0s
 => CACHED [3/8] COPY healthcheck.sh .                                                                                                                                                                                                           0.0s
 => CACHED [4/8] RUN chmod +x healthcheck.sh                                                                                                                                                                                                     0.0s
 => CACHED [5/8] RUN sed -i 's/\r$//' healthcheck.sh                                                                                                                                                                                             0.0s
 => CACHED [6/8] COPY configurator.sh .                                                                                                                                                                                                          0.0s
 => CACHED [7/8] RUN chmod +x configurator.sh                                                                                                                                                                                                    0.0s
 => CACHED [8/8] RUN sed -i 's/\r$//' configurator.sh                                                                                                                                                                                            0.0s
 => exporting to image                                                                                                                                                                                                                           0.0s
 => => exporting layers                                                                                                                                                                                                                          0.0s
 => => writing image sha256:214cde41ec12e39cb23a402379e72216b2c0dc9f7adbeaa527e41b46df4e60b6                                                                                                                                                     0.0s
 => => naming to docker.io/library/knightonline-sqlserver                                                                                                                                                                                        0.0s
 => resolving provenance for metadata file                                                                                                                                                                                                       0.0s
[+] Running 5/5
 ✔ knightonline-sqlserver                   Built                                                                                                                                                                                                0.0s 
 ✔ Network knightonline_backend             Created                                                                                                                                                                                              0.1s 
 ✔ Volume "knightonline_sqlserver_data"     Created                                                                                                                                                                                              0.0s 
 ✔ Volume "knightonline_sqlserver_scripts"  Created                                                                                                                                                                                              0.0s 
 ✔ Container knightonline-sqlserver-1       Started                                                                                                                                                                                              0.3s 
[+] Building 0.4s (36/36) FINISHED                                                                                                                                                                                                                    
 => [internal] load local bake definitions                                                                                                                                                                                                       0.0s
 => => reading from stdin 1.08kB                                                                                                                                                                                                                 0.0s
 => [sqlserver internal] load build definition from Dockerfile                                                                                                                                                                                   0.0s
 => => transferring dockerfile: 746B                                                                                                                                                                                                             0.0s
 => [kodb-util internal] load build definition from Dockerfile                                                                                                                                                                                   0.0s
 => => transferring dockerfile: 1.07kB                                                                                                                                                                                                           0.0s
 => [sqlserver internal] load metadata for mcr.microsoft.com/mssql/server:2022-latest                                                                                                                                                            0.0s
 => [sqlserver internal] load .dockerignore                                                                                                                                                                                                      0.0s
 => => transferring context: 2B                                                                                                                                                                                                                  0.0s
 => [kodb-util internal] load metadata for docker.io/library/golang:1.24-alpine                                                                                                                                                                  0.2s
 => [sqlserver 1/8] FROM mcr.microsoft.com/mssql/server:2022-latest                                                                                                                                                                              0.0s
 => [sqlserver internal] load build context                                                                                                                                                                                                      0.0s
 => => transferring context: 71B                                                                                                                                                                                                                 0.0s
 => CACHED [sqlserver 2/8] WORKDIR /var/opt/appscripts                                                                                                                                                                                           0.0s
 => CACHED [sqlserver 3/8] COPY healthcheck.sh .                                                                                                                                                                                                 0.0s
 => CACHED [sqlserver 4/8] RUN chmod +x healthcheck.sh                                                                                                                                                                                           0.0s
 => CACHED [sqlserver 5/8] RUN sed -i 's/\r$//' healthcheck.sh                                                                                                                                                                                   0.0s
 => CACHED [sqlserver 6/8] COPY configurator.sh .                                                                                                                                                                                                0.0s
 => CACHED [sqlserver 7/8] RUN chmod +x configurator.sh                                                                                                                                                                                          0.0s
 => CACHED [sqlserver 8/8] RUN sed -i 's/\r$//' configurator.sh                                                                                                                                                                                  0.0s
 => [sqlserver] exporting to image                                                                                                                                                                                                               0.0s
 => => exporting layers                                                                                                                                                                                                                          0.0s
 => => writing image sha256:214cde41ec12e39cb23a402379e72216b2c0dc9f7adbeaa527e41b46df4e60b6                                                                                                                                                     0.0s
 => => naming to docker.io/library/knightonline-sqlserver                                                                                                                                                                                        0.0s
 => [sqlserver] resolving provenance for metadata file                                                                                                                                                                                           0.0s
 => [kodb-util internal] load .dockerignore                                                                                                                                                                                                      0.0s
 => => transferring context: 2B                                                                                                                                                                                                                  0.0s
 => [kodb-util internal] load build context                                                                                                                                                                                                      0.0s
 => => transferring context: 77B                                                                                                                                                                                                                 0.0s
 => [kodb-util  1/15] FROM docker.io/library/golang:1.24-alpine@sha256:06545cc1ff10ddf04aebe20db1352ec7c96d1e43135767931c473557d0378202                                                                                                          0.0s
 => CACHED [kodb-util  2/15] RUN apk add --no-cache git ca-certificates                                                                                                                                                                          0.0s
 => CACHED [kodb-util  3/15] RUN apk add --no-cache gettext                                                                                                                                                                                      0.0s
 => CACHED [kodb-util  4/15] WORKDIR /var/lib/app                                                                                                                                                                                                0.0s
 => CACHED [kodb-util  5/15] RUN git --version                                                                                                                                                                                                   0.0s
 => CACHED [kodb-util  6/15] RUN git clone https://github.com/Open-KO/kodb-util.git                                                                                                                                                              0.0s
 => CACHED [kodb-util  7/15] WORKDIR /var/lib/app/kodb-util                                                                                                                                                                                      0.0s
 => CACHED [kodb-util  8/15] RUN git pull                                                                                                                                                                                                        0.0s
 => CACHED [kodb-util  9/15] RUN git submodule update --init --recursive --remote                                                                                                                                                                0.0s
 => CACHED [kodb-util 10/15] RUN go mod download                                                                                                                                                                                                 0.0s
 => CACHED [kodb-util 11/15] RUN set -e;     go build || (echo "Failed to build the kodb-util tool" && exit 1)                                                                                                                                   0.0s
 => CACHED [kodb-util 12/15] COPY kodb-util-config.yaml /var/lib/app/kodb-util/kodb-util-config.yaml.template                                                                                                                                    0.0s
 => CACHED [kodb-util 13/15] COPY cleanImport.sh /usr/local/bin/cleanImport.sh                                                                                                                                                                   0.0s
 => CACHED [kodb-util 14/15] RUN chmod +x /usr/local/bin/cleanImport.sh                                                                                                                                                                          0.0s
 => CACHED [kodb-util 15/15] RUN sed -i 's/\r$//' /usr/local/bin/cleanImport.sh                                                                                                                                                                  0.0s
 => [kodb-util] exporting to image                                                                                                                                                                                                               0.0s
 => => exporting layers                                                                                                                                                                                                                          0.0s
 => => writing image sha256:53ef2f4ed10377169f6240bb834232a5a93eed98c4e9533713239bc583a04bfe                                                                                                                                                     0.0s
 => => naming to docker.io/library/knightonline-kodb-util                                                                                                                                                                                        0.0s
 => [kodb-util] resolving provenance for metadata file                                                                                                                                                                                           0.0s
[+] Running 5/5
 ✔ knightonline-sqlserver                Built                                                                                                                                                                                                   0.0s 
 ✔ knightonline-kodb-util                Built                                                                                                                                                                                                   0.0s 
 ✔ Volume "knightonline_kodb_util_data"  Created                                                                                                                                                                                                 0.0s 
 ✔ Container knightonline-sqlserver-1    Healthy                                                                                                                                                                                                11.2s 
 ✔ Container knightonline-kodb-util-1    Started                                                                                                                                                                                                11.4s 
Installing a clean, up-to-date OpenKO database...
Already up to date.
------------------------------------------------------------------------------------------------------------------------
                                               OpenKO Database Utilities                                               
------------------------------------------------------------------------------------------------------------------------
Loading config...done
-- Clean --
Dropping KN_Online database...  Done
Dropping user knight...  Not found. Done
-- Import --
-- Importing databases --
databases successfully imported
-- Importing Schemas --
schemas successfully imported
-- Importing Users --
users successfully imported
-- Importing Logins --
logins successfully imported
-- Creating Tables --
table structures successfully created
-- Importing Table Data --
this may take several minutes
table data successfully imported in 8.07 seconds; batch size 16
-- Importing Views --
views successfully imported
-- Importing Stored Procedures --
stored procedures successfully imported

$ _
```

## Checklist
<!-- Go over all the following points, and put an `x` in all the boxes that apply -->

### DIFF
--- a/docker-compose.yaml
+++ b/docker-compose.yaml
@@ -36,7 +36,7 @@ services:
     healthcheck:
       test: ["CMD-SHELL", "/var/opt/appscripts/healthcheck.sh"]
       interval: 3s
-      retries: 10
+      retries: 30
       start_period: 10s
       timeout: 3s
   sqlserver.configurator:
--- a/docker/clean_setup.cmd
+++ b/docker/clean_setup.cmd
@@ -1,5 +1,13 @@
 :: Windows script
+:: Find script location on FS and use that as the working directory
+:: %~dp0: drive letter + back‑slash + directory
+set "SCRIPT_DIR=%~dp0"
+pushd "%SCRIPT_DIR%" || (
+    echo Failed to change to directory: "%SCRIPT_DIR%"
+    exit /b 1
+)
 cd ..
+echo Working directory set to: %CD%
 
 :: Removes any existing sqlserver/kodb-util images/volumes, 
 docker compose down --rmi all -v
--- a/docker/reset_database.cmd
+++ b/docker/reset_database.cmd
@@ -1,5 +1,13 @@
 :: Windows script
+:: Find script location on FS and use that as the working directory
+:: %~dp0: drive letter + back‑slash + directory
+set "SCRIPT_DIR=%~dp0"
+pushd "%SCRIPT_DIR%" || (
+    echo Failed to change to directory: "%SCRIPT_DIR%"
+    exit /b 1
+)
 cd ..
+echo Working directory set to: %CD%
 
 :: Performs a clean import of the OpenKO database.  This will 
 :: remove any existing OpenKO database instance matching the 
--- a/docker/resume_containers.cmd
+++ b/docker/resume_containers.cmd
@@ -1,5 +1,13 @@
 :: Windows script
+:: Find script location on FS and use that as the working directory
+:: %~dp0: drive letter + back‑slash + directory
+set "SCRIPT_DIR=%~dp0"
+pushd "%SCRIPT_DIR%" || (
+    echo Failed to change to directory: "%SCRIPT_DIR%"
+    exit /b 1
+)
 cd ..
+echo Working directory set to: %CD%
 
 :: Bring containers back online after a stop
 :: For first-time setup, or a clean database, use clean_setup.sh
--- a/docker/stop_containers.cmd
+++ b/docker/stop_containers.cmd
@@ -1,5 +1,13 @@
 :: Windows script
+:: Find script location on FS and use that as the working directory
+:: %~dp0: drive letter + back‑slash + directory
+set "SCRIPT_DIR=%~dp0"
+pushd "%SCRIPT_DIR%" || (
+    echo Failed to change to directory: "%SCRIPT_DIR%"
+    exit /b 1
+)
 cd ..
+echo Working directory set to: %CD%
 
 :: Stop the containers without removing any images or volumes
 :: resume with resume_containers.sh
--- a/docker/uninstall.cmd
+++ b/docker/uninstall.cmd
@@ -1,5 +1,13 @@
 :: Windows script
+:: Find script location on FS and use that as the working directory
+:: %~dp0: drive letter + back‑slash + directory
+set "SCRIPT_DIR=%~dp0"
+pushd "%SCRIPT_DIR%" || (
+    echo Failed to change to directory: "%SCRIPT_DIR%"
+    exit /b 1
+)
 cd ..
+echo Working directory set to: %CD%
 
 :: Removes any existing sqlserver/kodb-util images/volumes
 docker compose down --rmi all -v


docker-compose.yaml
```diff
@@ -0,0 +1,67 @@
...
+networks:
+  backend:
+
+# Use 'docker-compose down --rmi all -v' to completely remove database images.
+# This is useful for when you want to reset the database to the current
+# standard saved in the OpenKO-db project.
+services:
+  # Microsoft SQL Server service
+  sqlserver:
+    image: mcr.microsoft.com/mssql/server:2022-latest
```
stefannikolei: There is also another way. 

We could create an Image and publish it to the GitHub Container registry.  Then everyone could comsume it. And when he needs a new clean State then just throw away the container and volume. 

The publishing could be done in an GitHub Action. 


Otherwise this Looks good to me.

stefannikolei (approved):

.env
```diff
@@ -0,0 +1,5 @@
+# Docker environment configuration file
+# Do not commit/push changes to this file unless they
+# are changes/additions that everyone should recieve.
+Sa_Password=D0ckeIzKn!ght
```
twostars: Is this common practice to commit this at all? Should this perhaps be a `.default` or an `.example` file?
Ideally we'd still need a default if not set, and I'm not sure how we'd do that consistently as it's used by both `docker-compose.yaml` and `entry.sh`.

I'm just concerned about people committing this file at all and leaking their password in doing so, even if we aren't merging it in.

Same goes for `kodb-util-config.yaml`, which I feel might be slightly easier since it's loaded from script, you can just copy & rename the default file if it doesn't already exist.

docker-compose.yaml
```diff
@@ -0,0 +1,67 @@
...
+      - SA_PASSWORD=${Sa_Password}
+      - MSSQL_PID=Developer
+#      - MSSQL_TCP_ENCRYPTION=OFF
+    ports:
+      - 1433:1433
+    volumes:
+      - sqlserver_data:/var/opt/mssql
+    restart: always
+    healthcheck:
+      test: ["CMD-SHELL", "/opt/mssql-tools18/bin/sqlcmd -S sqlserver -U sa -P ${Sa_Password} -C -Q 'SELECT 1' || exit 1"]
```
twostars: Is this how we're intended to pass passwords? I'm just concerned about it (not) being escaped properly and such. For simple passwords it's fine at least, but just wondering if there's anything we can do -- or even need to do -- to ensure it's escaped properly here. Like for example (and again I don't know how this works) potentially breaking up each part into an argument here, like:
```
["CMD-SHELL", "/opt/mssql-tools18/bin/sqlcmd", "-S", "sqlserver", ...]
```
I don't know if that's how that even works but if it is, I'd assume that would ensure it's escaped correctly.

docker-compose.yaml
kenner2: The only thing I don't like about pulling this image (which would be a problem either way you grab it) is that it's a 500mb download every time you do a clean update.  I don't see a huge advantage to introducing a prebuilt image to a registry and having to manage it.  If anything, it might be best to rework the `refreshdb` process to be something that can be re-run whenever, so you can get a clean/updated database without having to redownload the sqlserver image

.env
kenner2: For container-level config, you can specify a default .env file and an optional override file.  That's what I tried to do initially.  However, for config available to the entire compose file, it will only recognize the .env file and there's no way to specify an override file.   It's very annoying because all the references in the compose file will get the .env value, and then within the container it will get the override value, causing potential mismatches.  So I cut that approach out; I can try playing with wrapping the sqlserver image in a Dockerfile and see if I can get all the parameters out of the compose file, but I don't know if that would actually work until I try it. 

I _can_ specify defaults such as:  ${Sa_Password:DefaultPasswordWhenBlank}.  I removed this because it was hiding issues loading from the .env files as I hashed all this out.  So for the .env file, I think it would be fine to do this as .env.default and put .env on .gitignore

For kodb-util-config.yaml, I think we should just leave it as is since I can't default it.  I'm not sure if I can do variable substitution from the .env file into this file - I'll have to check and try it out.  That would be a better solution since it won't require people to manually touch the files in any way to run the db.

docker-compose.yaml
kenner2: Good catch, I should definitely escape the variables within the shell scripts.  AFAIK the docker engine automatically handles escaping when using ${} within a docker-compose file, but I can try a few password combos to make sure.

kenner2: Most of these have been refactored such that it's the shell scripts referring to the .env variables.  I did test a password with characters that should break a script if not handled properly (`Br'Ea4"Z-R@`), and everything worked OK.

PR #662, opened 2025-12-12
Docker helper script tweaks

kenner2:
* Linux docker helper scripts: add export DOCKER_DEFAULT_PLATFORM=linux/amd64 to scripts to help support arm processors
* Windows docker helper scripts:  add popd to end of scripts to return to original directory

### DIFF
--- a/docker/resume_containers.sh
+++ b/docker/resume_containers.sh
@@ -29,4 +29,4 @@ echo "Working dir: " && pwd
 
 # Bring containers back online after a stop
 # For first-time setup, or a clean database, use clean_setup.sh
-docker compose up
+docker compose up -d
--- a/docker/resume_containers.cmd
+++ b/docker/resume_containers.cmd
@@ -11,7 +11,7 @@ echo Working directory set to: %CD%
 
 :: Bring containers back online after a stop
 :: For first-time setup, or a clean database, use clean_setup.sh
-docker compose up
+docker compose up -d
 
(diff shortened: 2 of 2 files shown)


docker/resume_containers.sh
```diff
@@ -1,4 +1,6 @@
 # Linux script
+export DOCKER_DEFAULT_PLATFORM=linux/amd64
```
stefannikolei: This is not needed here. It is only needed when creating the images

docker/resume_containers.sh
stefannikolei: This should be 'docker compose start' . Then the containers are started in the "background" 
Otherwise the stop_containers makes no sense.

docker/stop_containers.sh
```diff
@@ -1,4 +1,6 @@
 # Linux script
+export DOCKER_DEFAULT_PLATFORM=linux/amd64
```
stefannikolei: not needed for stopping

docker/reset_database.sh
```diff
@@ -1,4 +1,6 @@
 # Linux script
+export DOCKER_DEFAULT_PLATFORM=linux/amd64
```
stefannikolei: Also not needed. No new container is created.

stefannikolei (commented):
It works. Setting the env is only needed in the setup step when containers/images are created.

PR #679, opened 2025-12-13
Add a way to run Unit Tests

stefannikolei:


.github/workflows/tests_cmake.yml
```diff
@@ -0,0 +1,124 @@
+# This workflow uses actions that are not certified by GitHub.
+# They are provided by a third-party and are governed by
+# separate terms of service, privacy policy, and support
+# documentation.
+
+name: "Build server (CMake)"
```
twostars: This is two-fold really, but if this is separate, it'll need its own name.

Aside from that, I don't think this should be in a separate workflow; I think it should just be part of the regular workflow. The entire point of that is to test the builds in the first place, so it's just dispatching additional builds purely for the tests in this case.

Better to just build & run the tests in the original workflow.

.gitignore
```diff
@@ -62,3 +62,4 @@ build/
 
 # Do not ignore CMakeLists.txt
 !CMakeLists.txt
+cmake-build-debug
```
twostars: This doesn't really need to be in .gitignore. This is from CLion, which adds .gitignore to it locally so things shouldn't show up in the first place.

.gitmodules
```diff
@@ -63,3 +63,7 @@
 	path = deps/argparse
 	url = https://github.com/p-ranav/argparse
 	shallow = true
+[submodule "deps/googletest"]
+	path = deps/googletest
+	url = https://github.com/google/googletest.git
```
twostars: Drop the .git for consistency with the others

deps/googletest-cmake/CMakeLists.txt
```diff
@@ -0,0 +1,55 @@
...
+enable_testing()
+#
+# Include the gtest library. gtest_SOURCE_DIR is available due to
+# 'project(gtest)' above.
+include_directories(${gtest_SOURCE_DIR}/include ${gtest_SOURCE_DIR})
+
+########################################
+# Test files
+########################################
+file(GLOB TEST_SRC_FILES ${PROJECT_SOURCE_DIR}/tests/*.cpp)
```
twostars: I'd prefer to drop the globbing here; while it's convenient, it seems to also be a bit inconsistent.
It's probably better to just explicitly supply source files, then we know what we're getting every time.

Also: be sure to quote paths.

deps/googletest-cmake/CMakeLists.txt
```diff
@@ -0,0 +1,55 @@
...
+# option(BUILD_TESTS "Build all tests." ON) # Makes boolean 'test' available.
+
+# google test is a git submodule for the project, and it is also cmake-based
+#add_subdirectory(./deps/googletest)
+
+enable_testing()
+#
+# Include the gtest library. gtest_SOURCE_DIR is available due to
+# 'project(gtest)' above.
+include_directories(${gtest_SOURCE_DIR}/include ${gtest_SOURCE_DIR})
```
twostars: Be sure to quote paths.

CMakeLists.txt
```diff
@@ -114,3 +114,5 @@ if(OPENKO_BUILD_SERVERS)
   add_subdirectory(Server/ItemManager)
   add_subdirectory(Server/VersionManager)
 endif()
+
+add_subdirectory(deps/googletest-cmake)
```
twostars: Tests should have an option. I think it should perhaps be on by default, but I'm not sure.

.gitignore
stefannikolei: I added it because it was in my changes. will have a look..

stefannikolei: I checked it. The gitignore in .idea does not exclude this folder

twostars: There's no .gitignore in cmake-build-debug? It's CLion right, and it's up-to-date? That's so weird.

deps/googletest-cmake/CMakeLists.txt
stefannikolei: When we do not use globing this line can be removed and the files just added in the next line.

twostars (approved):

PR #687, opened 2025-12-16
Fix Regene Bugs

### LINKED ISSUE #629: CUser::Regene() bugs
CUser::Regene() (used on respawn) is a bit outdated and buggy.
Ignoring the functionality that doesn't exist, the 2 main problems with it right now are that:
1. The positioning logic is outdated and can fall through to not even finding a position (e.g. when dying in Moradon).
2. Doesn't restore HP. This is apparently just commented out; it shouldn't be.

stefannikolei:
Fixes #629

I hope I extracted the correct code from the posted reference.

Could not test it atm, because no windows machine to run the client on.

### DIFF
--- a/Server/Ebenezer/User.cpp
+++ b/Server/Ebenezer/User.cpp
@@ -2472,7 +2472,7 @@ void CUser::Regene(char* pBuf, int magicid)
 		m_fBlinkStartTime = TimeGet();
 //
 		m_bResHpType = USER_STANDING;
-//		HpChange( m_iMaxHp );
+		HpChange( m_iMaxHp );
 		m_bRegeneType = REGENE_NORMAL;
 	}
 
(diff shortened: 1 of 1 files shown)


Server/Ebenezer/User.cpp
```diff
@@ -2412,6 +2412,12 @@ void CUser::Regene(char* pBuf, int magicid)
 				}
 				else return;
 			}
+			int16_t spawnX = 0, spawnZ = 0;	// The point where you will be warped to.
```
twostars: This is overriding all of the previous blocks. I assume this was intended to be in an `else` block following them.

twostars (changes requested):

Server/Ebenezer/User.cpp
stefannikolei: Yep. Seems I missed the else in front of the return.

twostars (approved):

PR #705, opened 2025-12-20
Update to slnx

### LINKED ISSUE #704: Transition solutions to new .slnx format
.slnx is the new solution format, supported by up-to-date versions of VS2022, and VS2026 by default.
This is an XML format which is much easier to read and maintain than the old .sln format.

It would make dealing with dependencies (and verifying them) much simpler.

From what I've read, a simple solution upgrade *might* not preserve all the dependency chains and such properly, so it will take some time to go through and verify everything is transitioned correctly.

stefannikolei:
This upgrades all the solution files to the new Solution Format .slnx 
In this PR the old files are not deleted. 

This should be deleted in the near future. Otherwise they would run out of sync.

The Dependencies look good in my eyes. Omg is it easy to read now xD

Fixes #704


All.slnx
```diff
@@ -0,0 +1,203 @@
+<Solution>
+  <Configurations>
+    <Platform Name="Win32" />
+    <Platform Name="x64" />
+  </Configurations>
+  <Folder Name="/client tools/">
+    <Project Path="Client/KscViewer/KscViewer.vcxproj">
+      <BuildDependency Project="Client/JpegFile/JpegFile.vcxproj" />
+      <BuildDependency Project="FileIO/FileIO.vcxproj" />
```
twostars: This (FileIO) can be stripped. That's my fault, but it didn't actually end up depending on this.

All.slnx
```diff
@@ -0,0 +1,203 @@
...
+      <BuildDependency Project="deps/fetch-and-build-wrappers/fetch-spdlog.vcxproj" />
+      <BuildDependency Project="MathUtils/MathUtils.vcxproj" />
+      <BuildDependency Project="shared/shared.vcxproj" />
+    </Project>
+    <Project Path="shared/shared.vcxproj">
+      <BuildDependency Project="deps/fetch-and-build-wrappers/fetch-spdlog.vcxproj" />
+    </Project>
+  </Folder>
+  <Folder Name="/_deps/fetch wrappers/">
+    <Project Path="deps/fetch-and-build-wrappers/fetch-argparse.vcxproj" />
```
twostars: This should depend on sync-submodules

twostars (changes requested):
I can't really see much else other than that, for what exists.
What I do wonder, though, is can we pull in consistent project blocks into a solution, so we don't have to re-define their dependencies everywhere?

A lot of the hassle here is ensuring the dependency chains are setup consistently. I should imagine it should be able to import, though how the IDE would like that I don't know.

twostars (approved):

PR #721, opened 2025-12-28
Merge db-library into our main codebase

### LINKED ISSUE #720: Merge db-library into this repo
Its submodule isn't really needed. It's only used throughout this codebase, so there's no real reason for us to try and modularise this. It should be brought into the codebase directly.

twostars:
This really didn't need to be modularised into its own repo. This code can all be merged into our codebase as it's exclusively used here.

Resolves #720


stefannikolei:
Don't you want to include the history of it? I know it's only 10 commits..

PR #729, opened 2025-12-31
[Client] Fix the screen capture processes

selcukcukur:
## Pull request type

- [X] Bugfix
- [X] Feature
- [ ] Build related changes

Screenshot capture was always returning true, so it looked successful but no image was saved.

## What is the new behaviour?
Screenshots are now properly saved in the right format and orientation.

Removed the forced return, fixed the capture pipeline to use the correct DIB data, and flipped the image so it shows up correctly.

no, it was not used

## Checklist


twostars:
Please read our README regarding our policy on AI usage.

PR #738, opened 2026-01-01
Run format

stefannikolei:
I added this as a second branch to be able to evolve the formatting rules and keep the formatting in one commit

This enables us to exclude this commit from git blame by adding .git-blame-ignore-revs as a file to the repository


twostars:
Don't think we're necessarily changing the rules now so it's probably fine. I'll close this one.

PR #824, opened 2026-03-03
Implement ITEM_EXCHANGE evt functionality

### LINKED ISSUE #616: Server: Implement item exchanges (ITEM_EXCHANGE)

kenner2:
As part of #260 and #616 

Implements the following EXEC codes:

* RUN_EXCHANGE
* ~~CHECK_OLYMPIC_ACCOUNT~~
* ~~GIVE_OLYMPIC_ITEM~~

Implements the following LOGIC_ELSE codes:
* CHECK_EXCHANGE

## Pull request type
- [x] Feature

Currently the evt functionality for ITEM_EXCHANGE is not implemented, preventing events such as abyss gem exchange and Proconsul's Beginner Quests from functioning.

## What is the new behaviour?
ITEM_EXCHANGE cache and events implemented, allowing related events to function.

No

## Demo
Beginner Quest:
image
image

Abyss Gem:
image
First item pulled was an invalid ItemId:
image
Second item: Potion of Sagacity
Third Item: Priest Chitin
image

## Checklist


src/Server/Ebenezer/User.cpp
```diff
@@ -12486,6 +12506,173 @@ bool CUser::CheckNoExistItemAnd(const std::span<const ItemPair> items) const
...
+	// in their inventory.
+	if (exchange->RandomFlag != EXCHANGE_FLAG_STACK)
+		return openSlotCount > 0;
+
+	// RandomFlag == 0 exchanges are expected to grant all exchange items listed,
+	// so the inventory check is more robust
+	int slotsRequired = 0;
+	for (const int32_t iNum : exchange->ExchangeItemNumber)
+	{
+		const auto& item = m_pMain->m_ItemTableMap.GetData(iNum);
```
twostars: This shouldn't be a ref

src/Server/Ebenezer/User.cpp
```diff
@@ -12486,6 +12506,173 @@ bool CUser::CheckNoExistItemAnd(const std::span<const ItemPair> items) const
...
+
+		// acts as a bounds check on exchangeTable before populating it
+		if (sum > ONE_HUNDRED_PERCENT)
+		{
+			spdlog::error("User::RunExchange: Exchange configured for over 100% drop rate "
+						  "[charId={} exchangeId={} sum={}]",
+				m_pUserData->m_id, exchangeId, sum);
+			return;
+		}
+		if (sum < ONE_HUNDRED_PERCENT)
```
twostars, line 12611: This needs braces as it's spanning multiple lines (even though it's "one line" of code technically).

src/Server/Ebenezer/User.cpp
```diff
@@ -12653,17 +12840,79 @@ bool CUser::GiveItem(int itemid, int16_t count)
...
+	int sendIndex = 0;
+	SetByte(sendBuff, DB_OLYMPIC, sendIndex);
+	SetByte(sendBuff, OLYMPIC_GIVE_ITEM, sendIndex);
+	SetShort(sendBuff, _socketId, sendIndex);
+	SetString2(sendBuff, m_strAccountID, sendIndex);
+	SetDWORD(sendBuff, itemId, sendIndex);
+	SetShort(sendBuff, count, sendIndex);
+
+	int errorCode = m_pMain->m_LoggerSendQueue.PutData(sendBuff, sendIndex);
+	if (errorCode >= 10001)
```
twostars: This should have braces since it spans multiple linse. The errorCode here should also be using our constant. 

I realise this is 10001 (`SMQ_EMPTY`) but realistically this should be `>= SMQ_ERROR_RANGE`.

src/Server/Ebenezer/User.cpp
```diff
@@ -13573,7 +13822,7 @@ void CUser::ItemUpgrade(char* pBuf)
 	}
 	else
 	{
-		ItemLogToAgent(m_pUserData->m_id, "UPGRADE", ITEM_LOG_UPGRADE_FAILED, originItem.nSerialNum,
+		ItemLogToAgent(m_pUserData->m_id, "UPGRADE", ITEM_LOG_GIVE_ITEM, originItem.nSerialNum,
```
twostars: This doesn't look like it should've been changed.

src/shared/globals.h
```diff
@@ -100,6 +101,17 @@ enum e_Class : uint8_t
 	CLASS_UNKNOWN = 0xff
 };
 
+enum e_ExchangeFlag : uint8_t
+{
+	EXCHANGE_FLAG_STACK   = 0,
```
twostars: I don't know if I necessarily agree with this naming. It's a little confusing as to what's meant here, and we don't acknowledge 1-100 (which obviously don't really need constants, but still -- it's important).

We have:
0 - gives the lot (always), count = stack size
1-100 - gives a single item (rate is the the flag), count = stack size
101 - gives a single item (rate is the exchange "count"), count = percentage

I don't think "stack" (0) and "percent" (101) distinguish these cases when we take into consideration the other branch here.

That said I don't have better names as suggestions but I don't feel like these are sufficient.

src/Server/Ebenezer/User.cpp
```diff
@@ -12486,6 +12506,173 @@ bool CUser::CheckNoExistItemAnd(const std::span<const ItemPair> items) const
...
+		return false;
+
+	int16_t openSlotCount = 0;
+	for (int i = SLOT_MAX; i < INVENTORY_TOTAL; i++)
+	{
+		if (m_pUserData->m_sItemArray[i].nNum <= 0)
+			openSlotCount++;
+	}
+
+	// RandomFlag != 0 exchanges are expected to grant one random item.
```
twostars, line 12517: We should probably acknowledge what the actual official logic here was; the `sum(exchange->ExchangeItemCount[0..4]) > 9000` check, trying to determine if it's a percentage (vs stack sizes).

src/Server/Ebenezer/User.cpp
```diff
@@ -12486,6 +12506,173 @@ bool CUser::CheckNoExistItemAnd(const std::span<const ItemPair> items) const
...
+		spdlog::debug("User::RunExchange: User does not have all items [charId={} exchangeId={}]",
+			m_pUserData->m_id, exchangeId);
+		return;
+	}
+
+	// Reward processing varies by the exchange's RandomFlag
+	if (exchange->RandomFlag == EXCHANGE_FLAG_STACK)
+	{
+		if (!GiveItemAnd(exchangeItems))
+			spdlog::debug("User::RunExchange: Failed to grant all items [charId={} exchangeId={}]",
```
twostars, line 12588: Need braces here

src/Server/Ebenezer/User.cpp
```diff
@@ -12486,6 +12506,173 @@ bool CUser::CheckNoExistItemAnd(const std::span<const ItemPair> items) const
...
+				m_pUserData->m_id, exchangeId);
+
+		return;
+	}
+
+	if (exchange->RandomFlag >= EXCHANGE_FLAG_PERCENT)
```
twostars: This should just be `==`

src/Server/Ebenezer/EXEC.cpp
```diff
@@ -189,6 +195,18 @@ bool EXEC::Parse(const char* line, const std::string& filename, int lineNumber)
 			argsToParse = 1;
 			break;
 
+		// E CHECK_OLYMPIC_ACCOUNT {failedEventCount} {successfulEventCount}
```
twostars: These are event IDs not counts. The confusion is probably because I named then like `nEventNumFail` etc but they're specifically the event ID to run when it fails, and the event ID to run when it succeeds.

src/Server/Ebenezer/GameDefine.h
```diff
@@ -197,16 +197,34 @@ enum e_ItemType : uint8_t
...
-	ITEM_LOG_MERCHANT_SELL  = 2,
-	ITEM_LOG_MONSTER_GET    = 3,
-	ITEM_LOG_EXCHANGE_PUT   = 4,
-	ITEM_LOG_EXCHANGE_GET   = 5,
-	ITEM_LOG_DESTROY        = 6,
-	ITEM_LOG_WAREHOUSE_PUT  = 7,
-	ITEM_LOG_WAREHOUSE_GET  = 8,
-	ITEM_LOG_UPGRADE        = 9,
-	ITEM_LOG_UPGRADE_FAILED = 10
+	ITEM_LOG_MERCHANT_BUY           = 1,
```
twostars, line 201: I would be careful about these changes. These are what I roughly reversed but they're not necessarily accurate. They would need to be verified appropriately. They're just a really rough estimation pending going through each individually with purpose.

I wasn't being too strict about these matching exactly, since they end up just being meaningless numbers logged directly -- as numbers -- via a tool we also don't have a consistent official implementation of (there are 1 or 2 but they're wildly inconsistent, and yet still don't even deal with these types directly; they're just logged straight to file).

src/Server/Ebenezer/User.cpp
```diff
@@ -12653,17 +12840,79 @@ bool CUser::GiveItem(int itemid, int16_t count)
...
+	SetShort(sendBuff, count, sendIndex);
+
+	int errorCode = m_pMain->m_LoggerSendQueue.PutData(sendBuff, sendIndex);
+	if (errorCode >= 10001)
+		spdlog::error("User::GiveOlympicItem: Failed to send item [accountId={} itemId={} count={} "
+					  "errorCode={}]",
+			m_strAccountID, itemId, count, errorCode);
+}
+
+void CUser::CheckOlympicAccount(const int failedEventCount, const int successfulEventCount)
```
twostars: As before these are event IDs not counts.

Additionally, I don't know how I feel about considering these opcodes as being implemented when we're just sending the packet to Aujard and not handling them there. It feels half-implemented.

src/Server/Ebenezer/User.cpp
```diff
@@ -12653,17 +12840,79 @@ bool CUser::GiveItem(int itemid, int16_t count)
...
+	int sendIndex = 0;
+	SetByte(sendBuff, DB_OLYMPIC, sendIndex);
+	SetByte(sendBuff, OLYMPIC_CHECK_ACCOUNT, sendIndex);
+	SetShort(sendBuff, _socketId, sendIndex);
+	SetString2(sendBuff, m_strAccountID, sendIndex);
+	SetDWORD(sendBuff, failedEventCount, sendIndex);
+	SetDWORD(sendBuff, successfulEventCount, sendIndex);
+
+	int errorCode = m_pMain->m_LoggerSendQueue.PutData(sendBuff, sendIndex);
+	if (errorCode >= 10001)
```
twostars: Constant (SMQ_ERROR_RANGE) and braces

src/Server/Ebenezer/User.cpp
```diff
@@ -12375,6 +12391,10 @@ bool CUser::CheckWeight(int itemid, int16_t count) const
 
 bool CUser::CheckExistItem(int itemId, int16_t count) const
 {
+	// returning true for an invalid entry seems wrong, but is official
```
twostars: This should probably explain the context for this, in that it does it to allow for cases (like exchanges) to be ignored when those items are not set in a group. In that case it expects it to pass this check rather than fail (realistically it should just not bother to check it at all, but it's mgame, so).

src/Server/Ebenezer/User.cpp
kenner2: I corrected the enum name.  Officially this logs with log type 0, not 10.  We don't have an enum value for 0, though... I doubt it's anvil-specific, though.  I could add `ITEM_LOG_UNSPECIFIED` for the zero case and update this call.

src/Server/Ebenezer/User.cpp
kenner2: The Aujard side calls stored procedures that we do not have in the DB.  These events are never called in our .evt files - I think it was some one-off special server and they didn't remove the ops.  I only bothered to pull these in because the official exchange logic checks for it (but the code will never be hit given our EVTs).

I would be in favor of just pulling these out and descoping them.

src/Server/Ebenezer/User.cpp
twostars: That's because they use 0 for failed and 1 for success here. I wasn't caring too much about strictness here, since it's messy and it breaks the log types.

src/Server/Ebenezer/User.cpp
twostars: I mean yeah, we can just remove them entirely.

Issue #374, opened 2025-08-06
Missing green line under player name for clan leaders


koraykarakus:
Original

<img width="1366" height="768" alt="Image" src="https://github.com/user-attachments/assets/2d5ca4f0-c67b-429e-90bb-4493029d83d2" />

Current version 

<img width="1366" height="768" alt="Image" src="https://github.com/user-attachments/assets/868d7e7c-26ab-4fcc-9ef0-229c56cbcad4" />

twostars:
Two separate issues in one here. Bold was resolved by #377

koraykarakus:
Title of issue is edited.

About the unsolved problem:
d3dx9 does not support seperate colors for text and underline. This issue might be solved by implementing line draw inside CDFont class' drawtext function.

twostars:
You can just render the line directly. You're assuming it has to be associated with the font.
You don't need to modify CDFont for one (or well, technically 2) cases of a line that needs to be rendered.

PR #825, opened 2026-03-21
Green line under clan leaders name

reignsprime:
Draw the green line under the player name if they are a clan leader.

## Pull request type

- [x] Bugfix
- [ ] Feature
- [ ] Build related changes

No green line is drawn under the players name if they are a clan leader to address #374 

## What is the new behaviour?
The green line is drawn under the players name if they are a clan leader.

The packet was there to be read, however it was commented out in GameProcMain. 
Move e_KnightsDuty up to InfoPlayerOther so it could be seen by all players to draw the green line for all clan leaders.
I think virtual is the cleanest way to do this but open to suggestions.

No

## Demo
image

## Checklist

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -57,3 +57,5 @@ last-build-states/
 
 # Do not ignore CMakeLists.txt
 !CMakeLists.txt
+/deps/nanodbc
+/deps/zlib
--- a/.gitignore
+++ b/.gitignore
@@ -57,5 +57,3 @@ last-build-states/
 
 # Do not ignore CMakeLists.txt
 !CMakeLists.txt
-/deps/nanodbc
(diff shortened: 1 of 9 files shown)


reignsprime:
I dont use or understand clang and the errors it generates with code I havn't touched.

 I have tried to fix the clang with Claude but I don't understand it. I don't want to go down this path of using AI to work out clang.

I would rather just close my PR.

twostars:
clang-format just requires you to press CTRL+K, D (CRTL+K, then press D) to reformat the current entire source file/document.
clang-tidy usually just proposes the fixes to you already, but if there's specifics you don't understand, you can always just ask us.

PR #827, opened 2026-03-25
Improve asset directory identification

twostars:
Rather than duplicate, move core logic into AppThread for reuse. Additionally, rather than the default directory simply being up a level, first check the default directory exists in the current working directory, and then check if it exists in the parent directory.

Any other changes are purely intended to restructure for sharing the logic, it doesn't intend to behave any different.


src/shared/globals.h
```diff
@@ -9,8 +9,16 @@
...
+// Default location of server MAP directory, where .SMDs (server maps) reside.
+inline constexpr char DEFAULT_MAP_DIR[]              = "MAP";
+
+// Default location of server QUESTS directory, where Ebenezer's quest/event scripts (.EVTs) reside.
+inline constexpr char DEFAULT_QUESTS_DIR[]           = "QUESTS";
+
+// Default location of server event directory, where AI server's event scripts (.EVTs) reside.
+// This is the MAP directory by default.
+// NOTE: We have the AI server EVTs in the MAP dir by default. We only separate Ebenezer's EVTs into QUESTS.
+inline constexpr char DEFAULT_EVENT_DIR[]            = "MAP";
```
kenner2: Feels weird to have this as a separate var.  Good to have documented, but I don't think there's a good reason to have the string stored twice.  Is there a way we can do it as an alias/macro instead?

twostars: Not unless we swap them all to string_views, which we could do I guess

src/shared/globals.h
```diff
@@ -9,8 +9,16 @@
 
 #include <cstdint>
 
-inline constexpr char DEFAULT_MAP_DIR[]              = "../MAP/";
-inline constexpr char DEFAULT_QUESTS_DIR[]           = "../QUESTS/";
+// Default location of server MAP directory, where .SMDs (server maps) reside.
```
kenner2, line 13: Should all of these vars be in shared-server, instead of shared?

src/Server/shared-server/AppThread.h
```diff
@@ -88,6 +88,28 @@ class AppThread : public Thread
...
+	/// \brief Identifies an asset directory location, prioritising command-line > INI > defaults.
+	/// \param identifierName		Identifier name of a directory to be used for logging. e.g. "MAP"
+	/// \param commandLineDirectory	The directory passed to the command-line, if applicable. First in priority order.
+	/// \param configDirectory		The directory as stored in the INI config, if applicable. Second in priority order.
+	/// \param defaultDirectory		The default directory location, relative to the current working directory (e.g. ./MAP/).
+	///								As a fallback, will try the directory above (e.g. ../MAP/).
+	///								Last in the priority order.
+	/// \param outputPath			Path to output the final identified directory to.
+	/// \returns true when successful, false otherwise
+	AssetDirSource IdentifyAssetDir(const std::string_view identifierName,
```
kenner2: Should this be in a utility file instead of tied to the AppThread class?

src/shared/globals.h
twostars: There's no equivalent header currently and I think by itself, it's not really worth the hassle of introducing one. Maybe if there were more to shift in with it, but just for this I don't think it's worth

PR #829, opened 2026-04-04
UI Update - Hover Button Reset

### LINKED ISSUE #406: Button state is not reset when a closed UI element re opened
Opening UI and closing it causes button state to saved and when user opens it for the second time, button state is not started from default. In other words, on the second,third openings button seems as hovered. ( generally close button, did not see that behaviour on other buttons ).

reignsprime:
The UI hover button (X) state was not being reset when the window was being closed. 

## Pull request type

- [X] Bugfix
- [ ] Feature
- [ ] Build related changes

When you click X to close a window, the hover state is not being removed. When you open the window again, the hover state remains.

## What is the new behaviour?
When you click X to close a window it resets the state to normal. When the window is reopened it does not have the hover state.

To fix #406 
Change the called stated from on to normal. This was the bug causing the problem.

No

## Checklist

### DIFF
--- a/src/N3Base/N3UIButton.cpp
+++ b/src/N3Base/N3UIButton.cpp
@@ -192,7 +192,7 @@ uint32_t CN3UIButton::MouseProc(uint32_t dwFlags, const POINT& ptCur, const POIN
 			if (m_pParent != nullptr
 				&& UI_STATE_BUTTON_DOWN == m_eState) // 이전 상태가 버튼을 Down 상태이면
 			{
-				SetState(UI_STATE_BUTTON_NORMAL);        // 버튼을 On 상태로 만든다..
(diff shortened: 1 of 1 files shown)


twostars:
The approach here I'm not sure of. Being charitable, if this specific instance is actually a bug (and I double-checked; the state here is the same in current clients), then you're still only fixing one cause of this issue.

The official way they handle this, is they recurse through all the buttons on close/hide, resetting their states back to 'normal'.

As-is, yours will catch this specific case and reset it when the UI is shown again, but there's other reasons than just the explicit button click breaking the state here.

As such, I feel like the safest thing to do in this instance would just be to follow the offiical behaviour; reset all button states, rather than try to tweak this specific button's state. That way we can be sure the issue as a whole isn't a problem in *any* case, and we won't cause any unintended side-effects.

PR #831, opened 2026-04-07
Client - Clan leader line

reignsprime:
## Pull request type

- [X] Bugfix

The green clan leader line does not appear under the character name. Refer to #374.

## What is the new behaviour?
The green clan leader line appears if the character is a clan leader.

The packet was in the code, however it was commented out in GameProcMain.
Moved the character overhead information into CPlayerBase::RenderOverhead

No

## Demo
567234497-5d423a0d-3f2c-4b7e-a4c2-f170566a98d8

## Checklist

### DIFF
--- a/src/Client/WarFare/GameDef.h
+++ b/src/Client/WarFare/GameDef.h
@@ -419,18 +419,17 @@ struct __TABLE_PLAYER;
 
 struct __InfoPlayerOther
 {
-	int iFace;                  // Face type
-	int iHair;                  // Hair type
+	int iFace;             // Face type
+	int iHair;             // Hair type
 
-	int iCity;                  // Affiliated city
-	std::string szKnights;      // Clan name
-	int iKnightsGrade;          // Clan grade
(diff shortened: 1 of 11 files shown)


twostars:
This all looks fine enough, but I really don't like the `KnightsDuty()` virtual method. This seems... really hacky.

I think the better approach here, would be to have it behave slightly more like official where this logic ends up split up a little, so this stuff isn't running under `CPlayerBase::Render()` in the first place.

Officially, all of the overhead rendering ends up moved into its own virtual method anyway, which then means all related code gets implemented as needed per player type (NPC, other player, 'self').

I think this would probably be the better play here, to just shift it out and call it appropriately, even though it is a little bit more work to split it up. Then all related logic runs as appropriate, and not just the awkward `KnightsDuty()` call.

`KnightsDuty()` just seems excessive and hacky; short of splitting it up, it just being left in `__InfoPlayerBase` is more worth it IMO.

PR #832, opened 2026-04-07
Client - Initial Commit - Clan Cape

reignsprime:
## Pull request type

- [ ] Bugfix
- [X] Feature
- [ ] Build related changes

The clan cape/ cloak is not implemented in the client.
CUser::SendServerIndex is not implement in Ebenezer.

## What is the new behaviour?
The clan cloak is attached to the character. It displays the correct colour and patterns for the clan. It also displays the clan symbols. King cloak is also working.

This is my first iteration of the clan and king cloak. The cape is plugged into the shoulders. Most of the code already existed but it was commented out or not connected. The mesh is specific to each character model/ race. sCape is pulled from the knight table in the DB. This contains a code for the colour and the pattern. 

I have commented out the grid for debugging the clan cape mesh/ movement for the moment.

Setup CUser::SendServerIndex for the WIZ_SERVER_INDEX packet in Ebenezer
Setup the ServerIndex packet handling on the client side.
Setup GetSymbolFilename for the client side to get the image in the middle of the clan cloak.

Things still to do/ test:

## Screenshot
image

No

## Checklist

### DIFF
--- a/src/Client/WarFare/GameProcMain.cpp
+++ b/src/Client/WarFare/GameProcMain.cpp
@@ -2176,10 +2176,10 @@ bool CGameProcMain::MsgRecv_MyInfo_All(Packet& pkt)
 	// 기본값 읽기..
 	////////////////////////////////////////////////////////////
 
-	InitPlayerPosition(__Vector3(fX, fY, fZ)); // 플레이어 위치 초기화.. 일으켜 세우고, 기본동작을 취하게 한다.
+	InitPlayerPosition(__Vector3(fX, fY, fZ));                   // 플레이어 위치 초기화.. 일으켜 세우고, 기본동작을 취하게 한다.
 	s_pPlayer->RegenerateCollisionMesh();
-	s_pPlayer->AttachCloak(sCapeID);           // Cloak
-	s_pOPMgr->Release();                       // 다른 유저 관리 클래스 초기화..
+	s_pPlayer->AttachCloak(sCapeID, s_pPlayer->m_InfoExt.iRank); // Cloak
+	s_pOPMgr->Release();                                         // 다른 유저 관리 클래스 초기화..
 
 	if (m_pUICmdList != nullptr)
 		m_pUICmdList->CreateCategoryList();
@@ -2542,7 +2542,7 @@ bool CGameProcMain::MsgRecv_UserIn(Packet& pkt, bool bWithFX)
 	int iKnightsRank     = pkt.read<uint8_t>(); // 순위
 
 	int16_t sMarkVersion = pkt.read<int16_t>(); // Clan mark
-	int16_t sCapeID      = pkt.read<int16_t>(); // Clan cloak
+	int16_t sCapeID      = pkt.read<int16_t>(); // Cloak
 
 	int iLevel           = pkt.read<uint8_t>(); // 레벨...
 	e_Race eRace         = (e_Race) pkt.read<uint8_t>();
@@ -2574,7 +2574,7 @@ bool CGameProcMain::MsgRecv_UserIn(Packet& pkt, bool bWithFX)
 	/*uint8_t bInvisibilityType =*/pkt.read<uint8_t>();
 	/*int16_t sDirection        =*/pkt.read<int16_t>();
 	/*bool bIsChicken           =*/pkt.read<bool>();
-	/*uint8_t bRank             =*/pkt.read<uint8_t>();
+	uint8_t byRank = pkt.read<uint8_t>(); // Noble rank??
 	/*uint8_t bKnightsRank      =*/pkt.read<uint8_t>();
 	/*uint8_t bPersonalRank     =*/pkt.read<uint8_t>();
 
@@ -2624,8 +2624,9 @@ bool CGameProcMain::MsgRecv_UserIn(Packet& pkt, bool bWithFX)
 	pUPC->m_InfoBase.iAuthority = byAuthority;
 	pUPC->Init(eRace, iFace, iHair, dwItemIDs, iItemDurabilities, byItemFlags);
 	pUPC->RotateTo(DegreesToRadians(rand() % 360), true);
+	pUPC->m_InfoBase.iMarkVersion = sMarkVersion;
 	pUPC->KnightsInfoSet(iKnightsID, szKnightsName, iKnightsGrade, iKnightsRank);
-	pUPC->AttachCloak(sCapeID);
+	pUPC->AttachCloak(sCapeID, byRank);
 
 	//__KnightsInfoBase* pKIB = m_pUIKnightsOp->KnightsInfoFind(iKightsID);
 	//if(pKIB) pUPC->KnightsNameSet(pKIB->szName, 0xffff0000);
--- a/src/Client/WarFare/GameProcMain.cpp
+++ b/src/Client/WarFare/GameProcMain.cpp
@@ -1890,17 +1890,18 @@ bool CGameProcMain::MsgRecv_MyInfo_All(Packet& pkt)
 	/*int iAllianceID                       =*/pkt.read<int16_t>();
 	/*uint8_t byFlag                        =*/pkt.read<uint8_t>();
 
-	int iKnightNameLen = pkt.read<uint8_t>(); // 소속 기사단 이름 길이.
+	int iKnightNameLen = pkt.read<uint8_t>();                // 소속 기사단 이름 길이.
 	pkt.readString(szKnightsName, iKnightNameLen);
-	int iKnightsGrade = pkt.read<uint8_t>();  // 소속 기사단 등급
-	int iKnightsRank  = pkt.read<uint8_t>();  // 소속 기사단 순위
+	int iKnightsGrade = pkt.read<uint8_t>();                 // 소속 기사단 등급
+	int iKnightsRank  = pkt.read<uint8_t>();                 // 소속 기사단 순위
 
 	/*int16_t sMarkVersion            =*/pkt.read<int16_t>();
-	/*int16_t sCapeID                 =*/pkt.read<int16_t>();
+	int16_t sCapeID                   = pkt.read<int16_t>(); // Clan cape
 
 	// 기사단 관련 세팅..
 	s_pPlayer->m_InfoExt.eKnightsDuty = eKnightsDuty; // 기사단에서의 권한..
 	s_pPlayer->KnightsInfoSet(iKnightsID, szKnightsName, iKnightsGrade, iKnightsRank);
+	s_pPlayer->AttachCloak(sCapeID);
 	m_pUIVar->UpdateKnightsInfo();
 
 	s_pPlayer->m_InfoBase.iHPMax             = pkt.read<int16_t>();
@@ -2178,14 +2179,6 @@ bool CGameProcMain::MsgRecv_MyInfo_All(Packet& pkt)
 	InitPlayerPosition(__Vector3(fX, fY, fZ)); // 플레이어 위치 초기화.. 일으켜 세우고, 기본동작을 취하게 한다.
 	s_pPlayer->RegenerateCollisionMesh();
 
-	// berserk temp
-	//s_pPlayer->PlugSet(PLUG_POS_BACK, "item/babacloak.n3cplug_cloak", nullptr);	// 파트를 셋팅..
-	// end berserk temp
-
-	// berserk
-	//s_pPlayer->AttachCloak();
-
-	//..
 	s_pOPMgr->Release(); // 다른 유저 관리 클래스 초기화..
 
 	if (m_pUICmdList != nullptr)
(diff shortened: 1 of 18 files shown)


src/N3Base/N3Chr.cpp
```diff
@@ -1047,16 +1047,59 @@ CN3CPlug_Cloak::~CN3CPlug_Cloak()
...
 }
 
 bool CN3CPlug_Cloak::Load(File& file)
 {
-	CN3CPlugBase::Load(file);
+	// Read the name header (CN3BaseFileAccess::Load), then re-parse all
+	// CN3CPlugBase fields manually so we can redirect the mesh load to
+	// s_MngMesh (CN3Mesh) instead of PMeshSet (CN3PMesh). Cloak assets are
+	// .n3mesh files; loading them as .n3pmesh produces garbage vertex counts.
+	CN3BaseFileAccess::Load(file);
```
kenner2: Reviewing the ASM from the decompiled 1298 client, it does use `CN3CPlugBase::Load`; the ctidy comment is valid

src/Client/WarFare/GameProcMain.cpp
```diff
@@ -2178,14 +2179,6 @@ bool CGameProcMain::MsgRecv_MyInfo_All(Packet& pkt)
 	InitPlayerPosition(__Vector3(fX, fY, fZ)); // 플레이어 위치 초기화.. 일으켜 세우고, 기본동작을 취하게 한다.
 	s_pPlayer->RegenerateCollisionMesh();
 
-	// berserk temp
-	//s_pPlayer->PlugSet(PLUG_POS_BACK, "item/babacloak.n3cplug_cloak", nullptr);	// 파트를 셋팅..
```
kenner2, line 2182: I'm not sure where this particular block is sourced from ("babacloak" doesn't show up anywhere in the 1298 asm text search), so it should be safe to remove.  However, this is the correct place to call AttachCloak

src/Client/WarFare/PlayerBase.cpp
```diff
@@ -1849,6 +1868,44 @@ CN3CPlugBase* CPlayerBase::PlugSet(e_PlugPosition ePos, const std::string& szFN,
 	return pPlug;
 }
 
+void CPlayerBase::AttachCloak(int16_t sCapeID)
```
kenner2: This function (located around 0045D990 in the 1298 exe) takes an additional boolean arg, `isForce`.  

This gets passed into CPlayerBase::PlugSet (and by extension, CPlayerMyself::PlugSet).  CPlayerBase does nothing with this flag.  CPlayerMyself uses it in a transformation check:

```cpp
if (m_InfoBase.bIsTransformed && !force )
  {
    if ( pItemBasic )
      this->_.m_pItemPlugBasics[ePos] = pItemBasic;
    if ( pItemExt )
      this->_.m_pItemPlugExts[ePos] = pItemExt;
    return 0;
  }
```

The PlugSet funcs mentioned above need to have `bool isForce` added to their method signatures to support this.

src/Client/WarFare/PlayerBase.cpp
```diff
@@ -1849,6 +1868,44 @@ CN3CPlugBase* CPlayerBase::PlugSet(e_PlugPosition ePos, const std::string& szFN,
...
+{
+	m_Chr.CloakPlugSet("");
+
+	// -1 = no clan, 0 = no cape
+	if (sCapeID <= 0)
+		return;
+
```
kenner2, line 1886: Official func starts by checking `m_InfoBase.iRank == 1` (King) to see if the cape being loaded should be overridden with the King cape.

src/Client/WarFare/PlayerBase.cpp
```diff
@@ -1849,6 +1868,44 @@ CN3CPlugBase* CPlayerBase::PlugSet(e_PlugPosition ePos, const std::string& szFN,
 	return pPlug;
 }
 
+void CPlayerBase::AttachCloak(int16_t sCapeID)
+{
+	m_Chr.CloakPlugSet("");
+
+	// -1 = no clan, 0 = no cape
+	if (sCapeID <= 0)
```
kenner2: I don't think this is a valid check.  I don't see it in the official ASM. I assume the func needs to be called again when a user is kicked from clan, at which point you would want to unset the plug.

src/N3Base/N3Chr.h
```diff
@@ -471,6 +480,12 @@ class CN3Chr : public CN3TransformCollision
 		return m_Plugs[iIndex];
 	}
 
+	CN3CPlug_Cloak* CloakPlugSet(const std::string& szFN);
+	CN3CPlug_Cloak* CloakPlug()
```
kenner2, line 488: Is there a good reason we can't just use m_pCloakPlug directly?

src/N3Base/N3Chr.cpp
```diff
@@ -2201,6 +2253,24 @@ void CN3Chr::PlugAlloc(int iCount)
 	}
 }
 
+CN3CPlug_Cloak* CN3Chr::CloakPlugSet(const std::string& szFN)
```
kenner2: Use a more meaningful arg name, like `filename`

src/Client/WarFare/PlayerBase.cpp
```diff
@@ -1849,6 +1868,44 @@ CN3CPlugBase* CPlayerBase::PlugSet(e_PlugPosition ePos, const std::string& szFN,
...
+
+	if (pCloak->TexSet(szColour) == nullptr)
+		pCloak->TexSet("Item\\cloak_basic.dxt");
+
+	if (nPattern >= 1 && nPattern <= 5)
+		pCloak->TexOverlapSet(fmt::format("Item\\cloak_M_{:02}.dxt", nPattern));
+	else
+		pCloak->TexOverlapSet(static_cast<CN3Texture*>(nullptr));
+
+	pCloak->GetCloak()->Init(pCloak);
```
kenner2: Official function CN3Cloak::Init has 4 params:
CN3Mesh* mesh,
std::string textureFilename,
std::string symbolFilename,
std::string patternFilename (Item\\cloak_M_XX.dxt)

We can get the first two of those from `pCloak` within Init.  Init's function signature needs to be updated to accept symbolFilename and patternFilename.

kenner2: symbolFilename is populated via CGameProcedure::GetSymbolFilename, which is not currently implemented.  It'll need to be implemented for this.

The function declaration is:
`std::string CGameProcedure::GetSymbolFilename(const int serverIndex, const int knightsId, const int markVersion)`

The function does a string format: "symbol_us\\s_{}_{}_{}.dxt", passing in serverIndex, knightsId, and markVersion.  This might seem silly as a function but it's called in many places in the official exe

src/N3Base/N3Chr.cpp
twostars: Yeah, they're doing this because `CN3CPlugBase::Load()`'s only loading progressive meshes here.
But officially it loads both, which would avoid the need for this, and fix any other loaded plugs using the other meshes.

There's no guarantee that capes = old meshes. Though it's pretty common, it's not actually a hard and fast rule, so `CN3CPlugBase::Load()` should probably just be fixed instead.

src/Client/WarFare/PlayerBase.cpp
```diff
@@ -1849,6 +1868,44 @@ CN3CPlugBase* CPlayerBase::PlugSet(e_PlugPosition ePos, const std::string& szFN,
...
+
+	// Cloak colour and pattern
+	int nPattern          = sCapeID / 100;
+	int nColour           = sCapeID % 100;
+
+	std::string szColour;
+	if (nColour > 0)
+		szColour = fmt::format("Item\\cloak_c_{:02}.dxt", nColour);
+	else
+		szColour = "Item\\cloak_basic.dxt";
```
twostars: This fallback logic doesn't exist at all.

src/Client/WarFare/PlayerBase.cpp
```diff
@@ -1849,6 +1868,44 @@ CN3CPlugBase* CPlayerBase::PlugSet(e_PlugPosition ePos, const std::string& szFN,
...
+
+	if (pCloak->TexSet(szColour) == nullptr)
+		pCloak->TexSet("Item\\cloak_basic.dxt");
```
twostars: This fallback logic doesn't exist either.

src/N3Base/N3Chr.cpp
```diff
@@ -1047,16 +1047,59 @@ CN3CPlug_Cloak::~CN3CPlug_Cloak()
...
+		file.Read(szFN, nL);
+		szFN[nL] = '\0';
+		TexSet(szFN);
+	}
+
+	ReCalcMatrix();
+
+	if (m_pMesh == nullptr)
+		return false; // mesh file missing; CloakPlugSet will clean up
+	if (Tex() == nullptr)
```
twostars: This fallback logic also does not exist. If we were serious about it falling back if it absolutely had to support one, then it's probably best we define this as a constant, but I don't really think it's necessary.

It failing should just be a problem, not something we ignore IMO.

src/Client/WarFare/PlayerBase.cpp
reignsprime: Maybe I don't understand or have the wrong version of assets.

The ones I have include cloak_basic.dxt which was the simple dark blueish cloak you got when you first get the clan cloak. Then you had to go to the cape vendor and purchase a coloured/ patterned cloak.

image

image

twostars: So the thing is, it's just cape ID 0. This is the cape ID set here, and it goes through and uses that appropriately.
There's no overrides for this. I don't actually know what even uses the "basic" texture, it should just be both Cloak_C_00.dxt and (technically, even though it doesn't exist - but this one isn't a problem) Cloak_M_00.dxt in this case:

image

src/Client/WarFare/PlayerBase.cpp
twostars: Yeah this is actually not valid at all since 0 is a valid cape ID (the basic cape).
For no cape at all it's -1.

src/Client/WarFare/PlayerBase.cpp
```diff
@@ -1849,6 +1868,44 @@ CN3CPlugBase* CPlayerBase::PlugSet(e_PlugPosition ePos, const std::string& szFN,
...
+
+	// Attach cloak to shoulders
+	pCloak->m_nJointIndex = ITEM_POS_SHOULDER;
+
+	// Cloak colour and pattern
+	int nPattern          = sCapeID / 100;
+	int nColour           = sCapeID % 100;
+
+	std::string szColour;
+	if (nColour > 0)
```
twostars: This check is useless because cloak_c_00 is valid and expected for the basic cape anyway.

src/Client/WarFare/PlayerBase.cpp
```diff
@@ -1849,6 +1868,44 @@ CN3CPlugBase* CPlayerBase::PlugSet(e_PlugPosition ePos, const std::string& szFN,
...
+		szColour = fmt::format("Item\\cloak_c_{:02}.dxt", nColour);
+	else
+		szColour = "Item\\cloak_basic.dxt";
+
+	if (pCloak->TexSet(szColour) == nullptr)
+		pCloak->TexSet("Item\\cloak_basic.dxt");
+
+	if (nPattern >= 1 && nPattern <= 5)
```
twostars: I don't necessarily think there's a point to this check; it can pass 0, which technically will fail, but it'll end up emptying it anyway with it not existing. The arbitrary constraint here isn't really needed.

src/Client/WarFare/PlayerBase.cpp
reignsprime: Thanks, yep I see _C_00 that makes a lot more sense to use instead of basic.

src/Client/WarFare/PlayerBase.cpp
reignsprime: Thanks for the direction and assistance, it helped a lot.

Implemented CN3Cloak::Init and CGameProcedure::GetSymbolFilename.

src/Client/WarFare/PlayerBase.cpp
reignsprime: I had a look through the binary and was unable to locate the king cloak logic. I will implement it but wanted to check if the iRank should be in InfoBase or InfoExt. 

It is in InfoExt so I have left it there for the moment and passed it into the function.

No problem to move it all around, just need clarification on InfoBase or InfoExt.

twostars: It's in `__InfoPlayerBase` officially since it's used directly under `CPlayerBase`.

src/N3Base/N3Chr.cpp
reignsprime: I had a bit of a crack at the Load() this morning. I think it is right.

PR #842, opened 2026-05-29
plug glow fx fix

soL-KO:
## Pull request type

- [x] Bugfix
- [ ] Feature
- [ ] Build related changes

Currently, the item glow logic is hardcoded and incorrect. It indiscriminately applies a glow to items with an elemental damage value over 64. Not only this causes an issue where Type 5 items at the +7 falsely glow ( they should not be glowed at all ) but also it incorrectly applies a single, identical FX to all plugs with same elemental .

## What is the new behaviour?
The new behaviour accurately matches the original 1298 client executable's glow logic. It correctly distinguishes between different item types and applies the appropriate elemental FX based on the item's properties.

The `base` and `extension` and `fx` TBL files actually provide the necessary data structure to determine how the effects should be applied. Assuming familiarity with the original 1298 client's visual effects (for instance, rare items and upgrade items are supposed to have distinct FX), I refactored the glow logic to utilize the TBL data rather than relying on the incorrect hardcoded threshold.

TailFX is always be plus 1 that of the MainFX.
Plugs that have same elemental but different fx like Raptor and Glave should be overriden which is done by base table where dwEffectID2>0
Other than that it should always follow extension table EffectID which is dwIDK0 in codebase.

No.

## Demo

dark vane

## Checklist

### DIFF
--- a/src/Client/WarFare/GameDef.h
+++ b/src/Client/WarFare/GameDef.h
@@ -1376,6 +1376,7 @@ inline constexpr int FXID_WARP_ELMORAD            = 10047;
 inline constexpr int FXID_REGION_POISON           = 10100;
 inline constexpr int FXID_TARGET_POINTER          = 30001;
 inline constexpr int FXID_ZONE_POINTER            = 30002;
+inline constexpr int FXID_ELIXIRSTAFF_SPECIALCASE = 40000000;
 
 enum e_SkillMagicType4 : uint8_t
 {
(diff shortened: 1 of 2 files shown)


twostars:
Thanks for the submission.

This is close, but isn't actually fully accurate, particularly when using a staff. In this case, the base effect ID is offset based on what type of elemental damage is most dealt (fire - +0, ice - +10, lightning - +20).

The trace colour stuff doesn't matter too much officially (that's largely replaced; still exists, but it's largely worthless), but otherwise it does still exist somewhat with its >= 64 damage checks. Given it ends up being pretty useless I don't really care that much about preserving that side of it, but the effects themselves should at least behave accurately here.

soL-KO:
done

soL-KO:
we can also throw ron's staff fx in a case where race is a tie. But no real scenario in 1298

